PR #2, opened 2022-10-11
Add KHR_mesh_quantization

Eearslya:
Really all that's needed is to let fastgltf know about the extension.

The only changes it brings is allowing more types in accessors for mesh data.


spnda (approved):
I plan on adding a `validate` function to the glTF class that validates the parsed data more exactly as currently a lot of things are just taken as given (e.g. indices). That would then require support for the extension, but for now yeah looks good.

PR #3, opened 2022-10-12
Add material parameters and fix defaults

Eearslya:
Adds the `alphaMode`, `alphaCutoff`, and `doubleSided` params to `Material`.

Also fixes default values being missing when `KHR_texture_transform` is requested but fields are omitted.

### DIFF
--- a/src/fastgltf.cpp
+++ b/src/fastgltf.cpp
@@ -1135,9 +1135,9 @@ fg::Error fg::glTF::parseMaterials() {
 
         double alphaCutoff = 0.5;
         if (materialObject["alphaCutoff"].get_double().get(alphaCutoff) == SUCCESS) {
-            material.alphaCutoff = alphaCutoff;
+            material.alphaCutoff = static_cast<float>(alphaCutoff);
         } else {
-            material.alphaCutoff = 0.5;
+            material.alphaCutoff = 0.5f;
         }
 
         bool doubleSided = false;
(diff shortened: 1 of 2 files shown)


src/fastgltf.cpp
```diff
@@ -1118,6 +1118,35 @@ fg::Error fg::glTF::parseMaterials() {
...
+            } else {
+                return returnError(Error::InvalidGltf);
+            }
+        } else {
+            material.alphaMode = AlphaMode::Opaque;
+        }
+
+        double alphaCutoff = 0.5;
+        if (materialObject["alphaCutoff"].get_double().get(alphaCutoff) == SUCCESS) {
+            material.alphaCutoff = alphaCutoff;
```
spnda, line 1138: implicit double ->float conversion.

src/fastgltf.cpp
```diff
@@ -1529,14 +1558,13 @@ fg::Error fg::glTF::parseTextureObject(void* object, std::string_view key, Textu
...
 
-    if (!hasBit(this->extensions, Extensions::KHR_texture_transform)) {
-        info->rotation = 0.0f;
-        info->uvOffset = {0.0f, 0.0f};
-        info->uvScale = {1.0f, 1.0f};
-        return Error::None;
-    }
+    info->rotation = 0.0f;
+    info->uvOffset = {0.0f, 0.0f};
+    info->uvScale = {1.0f, 1.0f};
```
spnda, line 1567: Because you removed the check for the extension bits, the block below loading KHR_texture_transform will now execute even if the user did not enable the extension when creating the parser. Also, I would prefer that when the extension **is enabled** these defaults are only written below if the respective fields are not present.

spnda (changes requested):

PR #5, opened 2022-10-16
Change: Use simdjson's ondemand API

spnda:
In 0fe58d757627d0c3cac9ad36a5e60b29281483d9 we switched from on-demand -> DOM API as it was slightly faster. This switches back to the on-demand API while using a different approach to parsing to hopefully increase speeds over the current code using the DOM API.

### Design
The first iteration of the new parsing interface also helped with the DOM API, which is why that was merged separately already. However, this patch also tries to port that idea of iterating over all fields linearly in every function. Because currently, we only iterate over fields for the root of the JSON document and use hashed strings to speed up the switch. For every other function the order of fields may still be in the exact opposite (worse case) as how we read from them, which is why I want to use this idea everywhere in fastgltf.

### Performance

Performance has gotten worse which is why this is a draft PR and I will look into this drawback later on. From a quick profiling run it seems like the ondemand::parser::iterate function is taking a considerable amount of time while the rest executes within at most 500microseconds.

Mean vs Buggy gltf (16k lines, no embedded buffers or images) load from memory (3)

### DIFF
--- a/src/fastgltf.cpp
+++ b/src/fastgltf.cpp
@@ -193,7 +193,7 @@ namespace fastgltf {
 		}
 
 		std::uint64_t imageIndex;
-		if (sourceExtensionObject["source"].get_uint64().get(imageIndex) != SUCCESS) FASTGLTF_UNLIKELY {
+		if (sourceExtensionObject["source"].get(imageIndex) != SUCCESS) FASTGLTF_UNLIKELY {
 			return std::make_tuple(true, false, 0U);
 		}
 
@@ -282,13 +282,13 @@ namespace fastgltf {
 		using namespace simdjson;
 
 		std::uint64_t index;
-		if (object["index"].get_uint64().get(index) == SUCCESS) FASTGLTF_LIKELY {
+		if (object["index"].get(index) == SUCCESS) FASTGLTF_LIKELY {
 			info->textureIndex = static_cast<std::size_t>(index);
 		} else {
 			return Error::InvalidGltf;
 		}
 
-		if (auto error = object["texCoord"].get_uint64().get(index); error == SUCCESS) FASTGLTF_LIKELY {
+		if (auto error = object["texCoord"].get(index); error == SUCCESS) FASTGLTF_LIKELY {
 			info->texCoordIndex = static_cast<std::size_t>(index);
 		} else if (error != NO_SUCH_FIELD) FASTGLTF_UNLIKELY {
             return Error::InvalidJson;
@@ -296,14 +296,14 @@ namespace fastgltf {
 
 		if (type == TextureInfoType::NormalTexture) {
             double scale;
-			if (auto error = object["scale"].get_double().get(scale); error == SUCCESS) FASTGLTF_LIKELY {
+			if (auto error = object["scale"].get(scale); error == SUCCESS) FASTGLTF_LIKELY {
 				reinterpret_cast<NormalTextureInfo*>(info)->scale = static_cast<num>(scale);
 			} else if (error != NO_SUCH_FIELD) FASTGLTF_UNLIKELY {
                 return Error::InvalidGltf;
 			}
 		} else if (type == TextureInfoType::OcclusionTexture) {
 			double strength;
-			if (auto error = object["strength"].get_double().get(strength); error == SUCCESS) FASTGLTF_LIKELY {
+			if (auto error = object["strength"].get(strength); error == SUCCESS) FASTGLTF_LIKELY {
 				reinterpret_cast<OcclusionTextureInfo*>(info)->strength = static_cast<num>(strength);
 			} else if (error != NO_SUCH_FIELD) FASTGLTF_UNLIKELY {
                 return Error::InvalidGltf;
@@ -319,17 +319,17 @@ namespace fastgltf {
 				transform->uvOffset = {{ 0.0F, 0.0F }};
 				transform->uvScale = {{ 1.0F, 1.0F }};
 
-				if (textureTransform["texCoord"].get_uint64().get(index) == SUCCESS) FASTGLTF_LIKELY {
+				if (textureTransform["texCoord"].get(index) == SUCCESS) FASTGLTF_LIKELY {
 					transform->texCoordIndex = index;
 				}
 
 				double rotation = 0.0F;
-				if (textureTransform["rotation"].get_double().get(rotation) == SUCCESS) FASTGLTF_LIKELY {
+				if (textureTransform["rotation"].get(rotation) == SUCCESS) FASTGLTF_LIKELY {
 					transform->rotation = static_cast<num>(rotation);
 				}
 
 				ondemand::array array;
-				if (auto error = textureTransform["offset"].get_array().get(array); error == SUCCESS) FASTGLTF_LIKELY {
+				if (auto error = textureTransform["offset"].get(array); error == SUCCESS) FASTGLTF_LIKELY {
                     if (auto copyError = copyNumericalJsonArray<double>(array, transform->uvOffset); copyError != Error::None) {
                         return copyError;
                     }
@@ -337,7 +337,7 @@ namespace fastgltf {
                     return Error::InvalidGltf;
                 }
 
-				if (auto error = textureTransform["scale"].get_array().get(array); error == SUCCESS) FASTGLTF_LIKELY {
+				if (auto error = textureTransform["scale"].get(array); error == SUCCESS) FASTGLTF_LIKELY {
                     if (auto copyError = copyNumericalJsonArray<double>(array, transform->uvScale); copyError != Error::None) {
                         return copyError;
                     }
@@ -1239,7 +1239,7 @@ fg::Expected<fg::Asset> fg::Parser::parse(simdjson::ondemand::object root, Categ
         switch (hashedKey) {
             case force_consteval<crc32c("scene")>: {
                 std::uint64_t defaultScene;
-                if (object.value().get_uint64().get(defaultScene) != SUCCESS) FASTGLTF_UNLIKELY {
+                if (object.value().get(defaultScene) != SUCCESS) FASTGLTF_UNLIKELY {
                     return Expected<Asset>(Error::InvalidGltf);
                 }
                 asset.defaultScene = static_cast<std::size_t>(defaultScene);
@@ -1247,7 +1247,7 @@ fg::Expected<fg::Asset> fg::Parser::parse(simdjson::ondemand::object root, Categ
             }
             case force_consteval<crc32c("extensions")>: {
                 ondemand::object extensionsObject;
-                if (object.value().get_object().get(extensionsObject) != SUCCESS) FASTGLTF_UNLIKELY {
+                if (object.value().get(extensionsObject) != SUCCESS) FASTGLTF_UNLIKELY {
                     return Expected<Asset>(Error::InvalidGltf);
                 }
 
@@ -1261,7 +1261,7 @@ fg::Expected<fg::Asset> fg::Parser::parse(simdjson::ondemand::object root, Categ
                 }
 
                 ondemand::object assetObject;
-                if (object.value().get_object().get(assetObject) != SUCCESS) FASTGLTF_UNLIKELY {
+                if (object.value().get(assetObject) != SUCCESS) FASTGLTF_UNLIKELY {
                     return Expected<Asset>(Error::InvalidJson);
                 }
 
@@ -1400,11 +1400,15 @@ fg::Error fg::Parser::parseAccessors(simdjson::ondemand::array& accessors, Asset
 
         // Function for parsing the min and max arrays for accessors
         auto parseMinMax = [&](simdjson::ondemand::array& array, decltype(Accessor::max)& ref) -> fastgltf::Error {
-            decltype(Accessor::max) variant;
+			decltype(Accessor::max) variant;
 
             using double_vec = std::variant_alternative_t<1, decltype(Accessor::max)>;
             using int64_vec = std::variant_alternative_t<2, decltype(Accessor::max)>;
 
+            // It's possible the min/max fields come before the accessor type is declared, in which case we don't know the size.
+			// This single line is actually incredibly important as this function without it takes up roughly 15% of the entire
+			// parsing process on average due to vector resizing.
+            auto initialCount = accessor.type == AccessorType::Invalid ? 0 : getNumComponents(accessor.type);
             if (accessor.componentType == ComponentType::Float || accessor.componentType == ComponentType::Double) {
                 variant = FASTGLTF_CONSTRUCT_PMR_RESOURCE(double_vec, resourceAllocator.get(), 0);
             } else {
@@ -1475,7 +1479,7 @@ fg::Error fg::Parser::parseAccessors(simdjson::ondemand::array& accessors, Asset
             switch (hashedKey) {
                 case force_consteval<crc32c("bufferView")>: {
                     std::uint64_t bufferView;
-                    if (auto error = field.value().get_uint64().get(bufferView); error != SUCCESS) FASTGLTF_UNLIKELY {
+                    if (auto error = field.value().get(bufferView); error != SUCCESS) FASTGLTF_UNLIKELY {
                         return error == INCORRECT_TYPE ? Error::InvalidGltf : Error::InvalidJson;
                     }
                     accessor.bufferViewIndex = static_cast<std::size_t>(bufferView);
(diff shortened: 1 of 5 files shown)


spnda:
> 1. It is unsafe to assume that the builtin is fallback. On my laptop, it is arm64. It should not be necessary.
> 2. The `count_elements()` that the code keeps computing are probably harming.
> 3. Try compiling simdjson with SIMDJSON_VERBOSE_LOGGING and inspect the resulting log to see whether it looks sensible.
> 4. Doing some profiling to see whether you have specific bottlenecks could be great (e.g., perf record/perf report).

1. I've switched to include the simdjson header in fastgltf_parser.hpp. However, I really don't want to expose any headers that I don't actually need to expose. In this case, I use unique_ptr for anything from simdjson to essentially use PIMPL, which has worked amazingly so far. And as `simdjson::ondemand` is an alias for the builtin implementation, I can't just set it to anything as you've said. Is there any recommended way of forward declaring the stuff?
2. I've removed those, and by looking at some data from SIMDJSON_VERBOSE_LOGGING I noticed that it was fully reading the whole file searching for a field I was using to "verify" some basics of the file (asset info, extension info). I removed that part and I managed to reduce the runtime (for one specific glTF file) from 5.2ms average to 4.6ms average, which is a lot quicker but actually still slower than the 4.2ms with the DOM implementation.
3. See above. I saw a gazillion lines of "skip" before it even started reading anything.
4. I'm on my Windows machine right now, so I can't give much info. I did use Tracy once to analyze a few things in simdjson and found that stage1 was sometimes taking up significantly more than half the time (which seemed a bit suspect). But I will re-test on my Mac in a few days at most, so that I can get back to the simdjson issue with a bit more information on what takes the longest and is the biggest issue in my code.

Also sorry for the late response, I had a lot of stuff to do for school.

lemire:
> I'm on my Windows machine right now, so I can't give much info. 

If you are compiling under Windows, I recommend using Clang if possible. See

https://lemire.me/blog/2023/03/03/float-parsing-benchmark-regular-visual-studio-clangcl-and-linux-gcc/

spnda:
> If you are compiling under Windows, I recommend using Clang if possible. See
> 
> https://lemire.me/blog/2023/03/03/float-parsing-benchmark-regular-visual-studio-clangcl-and-linux-gcc/

Thanks for the read. I've gone ahead and I've just run my benchmarks using Clang-CL 16 (I used clang-cl from the LLVM downloads). It has actually increased the runtime speed of the fastgltf cases by often fourfold with both master and this branch. Though this PR's branch is still lacking behind consistently.

image

lemire:
> It has actually increased the runtime speed of the fastgltf cases by often fourfold with both master and this branch.

I see something like a factor of two... but yeah, Visual Studio has disappointing performance in some cases I care about.

spnda:
@lemire I've revisited this PR because I want to add support for #41, which I think I want to do using raw JSON tokens, as provided by the on demand API.

Using verbose logging I already managed to restructure my code to have no skips at all during the entire parsing process. I've looked through some things using Xcode Instruments, and have found some issues. I fixed some of the issues with allocations and vector resizing where it wasn't needed, but this one stumps me (there's more than just the preview shown, and ignore the return at the top of the function I accidentally committed that):
https://github.com/spnda/fastgltf/blob/fc5b3c2e9a105bed89332f6187b3d3d41a4a69eb/src/fastgltf.cpp#L1330-L1400

This lambda function takes up 36% of the entire average runtime of the parseAccessors function. It essentially just reads a number from an array, and then checks if it's a double, float, or integer and then converts them appropriately. Is there some better way to do this with simdjson, so that its runtime is not so excessive? Do you have any suggestions on improving that function?

Screenshot 2024-02-02 at 15 26 25

lemire:
@spnda Having a look.

src/fastgltf.cpp
```diff
@@ -204,90 +204,97 @@ namespace fastgltf {
 		return Error::InvalidJson;
 	}
 
+    template <typename X, typename T, std::size_t N>
+    [[nodiscard, gnu::always_inline]] Error copyNumericalJsonArray(simdjson::ondemand::array& array, std::array<T, N>& target) {
```
lemire, line 208:
```suggestion
    [[nodiscard, gnu::always_inline]] inline Error copyNumericalJsonArray(simdjson::ondemand::array& array, std::array<T, N>& target) {
```

lemire: GCC won't inline the function unless you mark it as inline.

src/fastgltf.cpp
```diff
@@ -1303,366 +1313,445 @@ fg::Expected<fg::Asset> fg::Parser::parse(simdjson::dom::object root, Category c
...
+                            (*doubles).emplace_back(static_cast<double>(value));
+                        } else if (auto* ints = std::get_if<int64_vec>(&variant); ints != nullptr) {
+                            (*ints).emplace_back(static_cast<std::int64_t>(value));
+                        } else {
+                            return Error::InvalidGltf;
+                        }
+                        break;
+                    }
+                    case ondemand::number_type::unsigned_integer: {
+                        // Note that the glTF spec doesn't care about any integer larger than 32-bits, so
```
lemire, line 1381: It should only trigger `unsigned_integer` if the integer is larger or equal to 2**63, so maybe this case is in error for you, or needs to be treated as a double?

src/fastgltf.cpp
```diff
@@ -1303,366 +1313,445 @@ fg::Expected<fg::Asset> fg::Parser::parse(simdjson::dom::object root, Category c
...
-                decltype(Accessor::max) variant;
+            using double_vec = std::variant_alternative_t<1, decltype(Accessor::max)>;
+            using int64_vec = std::variant_alternative_t<2, decltype(Accessor::max)>;
 
-				using double_vec = std::variant_alternative_t<1, decltype(Accessor::max)>;
-				using int64_vec = std::variant_alternative_t<2, decltype(Accessor::max)>;
+            // It's possible the min/max fields come before the accessor type is declared, in which case we don't know the size.
+			// This single line is actually incredibly important as this function without it takes up roughly 15% of the entire
+			// parsing process on average due to vector resizing.
+            auto initialCount = accessor.type == AccessorType::Invalid ? 0 : getNumComponents(accessor.type);
```
lemire, line 1341: Is initialCount used anywhere?

src/fastgltf.cpp
```diff
@@ -1303,366 +1313,445 @@ fg::Expected<fg::Asset> fg::Parser::parse(simdjson::dom::object root, Category c
...
-        // byteOffset is optional, but defaults to 0
-        std::uint64_t byteOffset;
-        if (auto error = accessorObject["byteOffset"].get_uint64().get(byteOffset); error == SUCCESS) {
-            accessor.byteOffset = static_cast<std::size_t>(byteOffset);
-        } else if (error != NO_SUCH_FIELD) {
-            return Error::InvalidGltf;
-        }
+        // Function for parsing the min and max arrays for accessors
+        auto parseMinMax = [&](simdjson::ondemand::array& array, decltype(Accessor::max)& ref) -> fastgltf::Error {
+			return Error::None;
```
lemire, line 1332: So the first line of the function is a return?

src/fastgltf.cpp
spnda: The data included in the arrays can at most be 2\*\*32 (or 32-bit floats), so that that case is just never reached probably. If it's only ever taken with values larger or equal to 2**63 I guess I would be safe to just remove it. Though for correctness, would it really be bad to keep it?

src/fastgltf.cpp
spnda: yeah, oops. initialCount should be used two lines further down for reserving the vector capacity:
```cpp
auto vec = FASTGLTF_CONSTRUCT_PMR_RESOURCE(double_vec, resourceAllocator.get(), 0);
vec.reserve(initialCount);
variant = std::move(vec);
```

src/fastgltf.cpp
spnda: In my original comment I mentioned that I committed some things I didn't exactly intend. I'll update the branch in a second with a few fixes.

src/fastgltf.cpp
```diff
@@ -1303,366 +1313,449 @@ fg::Expected<fg::Asset> fg::Parser::parse(simdjson::dom::object root, Category c
...
+                auto vec = FASTGLTF_CONSTRUCT_PMR_RESOURCE(double_vec, resourceAllocator.get(), 0);
+				vec.reserve(initialCount);
+				variant = std::move(vec);
+            } else {
+                auto vec = FASTGLTF_CONSTRUCT_PMR_RESOURCE(int64_vec, resourceAllocator.get(), 0);
+				vec.reserve(initialCount);
+				variant = std::move(vec);
+            }
 
+			for (auto element : array) {
```
lemire, line 1352: I am not sure I understand this code, but I would expect that you could do...

```
if(accessor.componentType == ComponentType::Double) {
  auto v = std::get<double_vec>(variant);
  for (auto element : array) {
    // error handing as needed
    v.push_back(element.get_double());
  }
}
```

And so forth. The number type is a fallback for complicated scenarios, where you need to branch according to the number type, but if you know that a double is good enough in some instances, then just grab that.

spnda: Okay, so a few things first:
- I should actually be down casting to single precision in all cases:
    > For floating-point components, JSON-stored minimum and maximum values represent single precision floats and SHOULD be rounded to single precision before usage to avoid any potential boundary mismatches.
- JSON does not make the distinction between integers and floating point numbers, as you know. Therefore, based on `accessor.componentType`, I have to cast either to float or to an integer type. The glTF spec specifically says this:
    > Array elements MUST be treated as having the same data type as accessor’s componentType.
    
    From that, I understand that if the component type specifies an integer but the value is `3.5` for whatever reason in the JSON, I have to just cast that to an integer. The specification makes no further clarifications on how the values should be treated as far as I can tell.
    
As I don't know how simdjson parses or identifies the numbers internally, would you say it'd be safe to just have this instead to avoid the `ondemand::number` object?
```cpp
switch (accessor.componentType) {
    case ComponentType::Double:
        v.emplace_back(element.get_double();
        break;
    case ComponentType::Float:
        ...
}
```

lemire:
> This lambda function takes up 36% of the entire average runtime of the parseAccessors function. It essentially just reads a number from an array, and then checks if it's a double, float, or integer and then converts them appropriately. Is there some better way to do this with simdjson, so that its runtime is not so excessive? Do you have any suggestions on improving that function?


The whole function does a lot of complicated work.

If you turn it into something like ...

```
  auto v = std::get<double_vec>(variant);
  for (auto element : array) {
    // error handing as needed
    v.push_back(element.get_double());
  }
```

You should be pretty close to be limited by the number parsing speed. If not, then you have other overhead.

My recommendation is to try to break down your code into simpler functions. There is a lot going on.


It may help also to include a standard benchmark as part of the project. If I could run a benchmark easily, I might be able to give you more insights. Just looking at high level code buys only so much. I need to know what the code is doing in practice.

lemire:
This part is important: I recommend including a standard benchmark as part of your project. If I could just do...

```
cmake --build build
./build/benchmark
```

That would tremendously useful. Not just to me but to anyone who cares about performance.

lemire:
Feel free to steal some of my benchmarking code:

https://github.com/lemire/Code-used-on-Daniel-Lemire-s-blog/tree/master/2023/11/28/benchmarks

src/fastgltf.cpp
lemire: Converting to float from double is typically done with a single instruction (e.g., cvtsd2ss). It is a high latency operation, but not overly expensive. So if you don't need to immediately check the result, it should be quite cheap. 

> would you say it'd be safe to just have this 

I wouldn't use a switch case if I could avoid it to handle individual elements. I would have distinct code paths based on `componentType`. It depends on the context, with repeated calls to a switch/case can be detrimental to the performance. Not always, it depends on how the compiler decides to handle it and how precisely you wrote the code, but to be safe, assuming that a switch/case adds significant overhead because it might. It is not the switch case per se that's the problem, it is that it prevents some useful optimizations by the compiler.

Of course, trying to parse `"3.5"` as an integer will cause an error, but you could handle this with an exceptional case (using a cold path if needed).

lemire: > So the proposed code isn't a possibility

My proposal is that you try it and benchmark. I submit to you that you need to run experiments to find the right design, even if it means temporarily having code that does not solve the right problem. You need to identify the bottleneck and that's damn difficult if you are just starting at high-level code.

src/fastgltf.cpp
lemire: > would it really be bad to keep it?

I recommend that you simplify your code prior to optimizing it. It is much easier to optimize simpler code.

I tried dumping the assembly from this function and I got several pages of instructions. It is just not possible to understand that much code at once. If you can trim it down to something simpler, then it might be much easier to understand where the performance challenges are.

src/fastgltf.cpp
lemire: Note that I don't know whether it should be inlined, this was just a side remark. I have not benchmarked.

lemire:
My recommendation is to rely more on experiments and repeated high-quality benchmarks, than on profiling. Profilers mislead us all the time. They may point at a potential direction, but that's all. There are exceptions. If you do pure numerical analysis, you may find 3 lines of code where your code is spending 99% of its time, and then you rewrite them in assembly and you are done... but for a problem such as this one, experiments are very likely to be the way forward. One needs to gain insights into what is costing what.

For example, how much does the get_number() function costs as opposed to get_double()? This takes just one experiment to find it out.

spnda:
> This part is important: I recommend including a standard benchmark as part of your project. If I could just do...
> 
> ```
> cmake --build build
> ./build/benchmark
> ```
> 
> That would tremendously useful. Not just to me but to anyone who cares about performance.

I already include benchmarks (which I am also using right now locally) which use the Catch2 testing framework. I should probably add something about how to build tests and run them in the README...

Though this is how you'd build and run the benchmarks. Note that these do require external downloads of sample assets because I currently just bench an entire run over a JSON asset. I mostly just wrote them to work well for me locally and for the CI.
```
python fetch_test_deps.py
cmake -B build -DFASTGLTF_ENABLE_TESTS=ON
cmake --build build --target fastgltf_tests
build/tests/fastgltf_tests "[gltf-benchmark]"
```
(note that the quotation marks in the last line might just be a requirement of zsh which I'm using locally, so you might have to remove those).

src/fastgltf.cpp
spnda: I wrote a small little benchmark which tried to parse the same JSON array which contains integers and doubles (50-50 distribution). Turns out `get_number` is ever so slightly faster in my little experiment, even with the switch-case covering those three cases.

lemire:
> I already include benchmarks

The parseAccessors is tiny component of this benchmark.

Capture d’écran, le 2024-02-02 à 16 40 02

Capture d’écran, le 2024-02-02 à 16 41 23

lemire:
@spnda Here is my recommandation.

Take your current code, do all the parsing and error handling, but do not touch your data structures. Do not call reserve, emplace_back, push_back at least as far as the numbers are concerned.

Next do the reverse. Do not call get_number... iterate through the array, but do not parse the values. Just add random numbers in your data structures (say random numbers between 0 and 1).

And then compare with the current code.

Lay out the three results. This should tell  you a lot about where the time goes.

My generally point is that you need to run experiments, and collect data.

spnda:
> @spnda Here is my recommandation.
> 
> Take your current code, do all the parsing and error handling, but do not touch your data structures. Do not call reserve, emplace_back, push_back at least as far as the numbers are concerned.
> 
> Next do the reverse. Do not call get_number... iterate through the array, but do not parse the values. Just add random numbers in your data structures (say random numbers between 0 and 1).
> 
> And then compare with the current code.
> 
> Lay out the three results. This should tell you a lot about where the time goes.
> 
> My generally point is that you need to run experiments, and collect data.

Well, big thanks for taking the time to look over my project. I'll do some testing along the lines of what you suggested. I'm not knowledgeable about these low-level optimizations, so much of this is still new to me. Using profiling I managed to find a few issues in my code apart from the `parseMinMax` lambda, which has already had a big impact on the ultimate speed. The original question, of why DOM was faster, is mostly solved actually, or at least on my arm64 M3. I still expect the ondemand code to be quicker than it is currently, but any improvement is welcome right now. And on my 5800X I still have a measurable performance regression in all cases. I guess I'll just have to do more investigation on both chips at the same time to see what differences there are to cover both. Another user had also suggested on Discord that the slowdown might be cache/memory related, which I have yet to test/confirm.

lemire:
@spnda The reason I raise these questions is that software performance is complex and requires hard numbers and experimentation.

For example, just by tracing the code without too much attention, I noticed that you are often enough storing a single integer in a std::vector which is itself a std::variant. This is fine if that's what you need, but constructing these data structures is expensive. And parsing routine themselves may not carry much weight.

But it is impossible to tell without experimentations and profiling won't tell you.

PR #14, opened 2023-03-29
Android file load

DethRaid:
This PR adds functionality to load files from within an Android APK

Note: This change only supports loading files with embedded buffers and textures. External buffers and textures will require more work. I'll implement them after spnda's URI changes make it in

### DIFF
--- a/src/fastgltf.cpp
+++ b/src/fastgltf.cpp
@@ -2185,7 +2185,7 @@ bool fg::GltfDataBuffer::loadFromFile(const fs::path& path, std::uint64_t byteOf
 
     // Skip over as many characters as possible until EOF, then get how many characters were skipped.
     file.ignore(std::numeric_limits<std::streamsize>::max());
-    auto length = file.gcount() - byteOffset;
+    auto length = file.gcount();
     if (length == 0 || file.bad())
         return false;
 
--- a/src/fastgltf.cpp
+++ b/src/fastgltf.cpp
@@ -357,6 +357,17 @@ fg::Error fg::glTF::validate() {
         if (accessor.bufferViewIndex.has_value() &&
             accessor.bufferViewIndex.value() >= parsedAsset->bufferViews.size())
             return Error::InvalidGltf;
+
+        if (!std::holds_alternative<std::monostate>(accessor.max)) {
+            if ((accessor.componentType == ComponentType::Float || accessor.componentType == ComponentType::Double)
+                && !std::holds_alternative<std::vector<double>>(accessor.max))
+                return Error::InvalidGltf;
+        }
+        if (!std::holds_alternative<std::monostate>(accessor.min)) {
+            if ((accessor.componentType == ComponentType::Float || accessor.componentType == ComponentType::Double)
+                && !std::holds_alternative<std::vector<double>>(accessor.min))
+                return Error::InvalidGltf;
+        }
     }
 
     for (const auto& animation : parsedAsset->animations) {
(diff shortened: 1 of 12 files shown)


src/fastgltf.cpp
```diff
@@ -2203,6 +2251,7 @@ bool fg::GltfDataBuffer::loadFromFile(const fs::path& path, std::uint64_t byteOf
     file.read(reinterpret_cast<char*>(bufferPointer), static_cast<std::streamsize>(dataSize));
     std::memset(bufferPointer + dataSize, 0, allocatedSize - dataSize);
     return true;
+    #endif
```
spnda, line 2254: I strongly thing this is wrong. What if you are on Android but want to use std::ifstream for whatever reason? Also, one could call the default constructor and then the asset manager would be nullptr, causing a segfault here.

Instead, I think the `loadFromFile_Android` should be called `loadFromFileAndroid` and be public, so that its use is explicit. (And that should have an `assert(asset_manager != nullptr)`

DethRaid: I've moved the Android loading functionality to a child class. The only way to construct it is with an `AAssetManager`. This will help prevent people from loading an Android asset without an `AAssetManager`

I've renamed the new method to `loadFromAndroidAsset` since technically the assets in the APK aren't files. You can still use `loadFromFile` to load a glTF from a file external to your APK

spnda: Looks good. Then I'd guess `AndroidGltfDataBuffer` should make `loadFromFile` private, because it wouldn't make sense to have those functions available to a struct made exclusively to load APK assets? So, effectively this:
```cpp
class AndroidGltfDataBuffer : public GltfDataBuffer {
private:
    using GltfDataBuffer::loadFromFile;
}
```

src/fastgltf.cpp
```diff
@@ -2181,8 +2181,51 @@ bool fg::GltfDataBuffer::copyBytes(const std::uint8_t* bytes, std::size_t byteCo
...
+#pragma endregion
+
+#pragma region AndroidGltfDataBuffer
+#if defined(__ANDROID__)
+fg::AndroidGltfDataBuffer::AndroidGltfDataBuffer(AAssetManager* asset_manager_in) noexcept : asset_manager{asset_manager_in} {}
+
+bool fg::AndroidGltfDataBuffer::loadFromAndroidAsset(const fs::path& path, std::uint64_t byteOffset) noexcept {
+    if (asset_manager == nullptr) {
+        return false;
+    }
```
spnda, line 2227: Shouldn't this be a exception? (Or assert, but I would dislike asserts because those should only be used for checks of internal state, not something the user could break)

src/fastgltf_parser.hpp
```diff
@@ -321,14 +313,29 @@ namespace fastgltf {
...
         /**
          * Returns the size, in bytes,
          * @return
          */
         [[nodiscard]] inline std::size_t getBufferSize() const noexcept;
     };
 
+    #if defined(__ANDROID__)
+    class AndroidGltfDataBuffer : public GltfDataBuffer {
+        AAssetManager* asset_manager = nullptr;
```
spnda, line 326: This is a redundant initializer given that the only constructor of `AndroidGltfDataBuffer` writes a value to `asset_manager`. Btw, lowerCamelCase please?

src/fastgltf.cpp
DethRaid: I wasn't sure if you were exception-friendly

src/fastgltf.cpp
DethRaid: I don't think it should be private. On Android, you can still load a file from outside the APK using standard C++ classes

src/fastgltf.cpp
spnda: Right not, no. But I am thinking of changing that to some core parts of the library because simdjson by default usually also uses exceptions. I can imagine that using exceptions could save a few things here and there. I guess it's fine as is and I will change it myself later if I do switch to exceptions.

src/fastgltf.cpp
spnda: In that case you can just use `GltfDataBuffer`, no?

DethRaid: I don't think so. I wanted to make a subclass so that you could only try to load an Android asset if you've supplied an `AAssetManager`

src/fastgltf.cpp
```diff
@@ -2210,6 +2214,56 @@ std::size_t fg::GltfDataBuffer::getBufferSize() const noexcept {
...
+bool fg::AndroidGltfDataBuffer::loadFromAndroidAsset(const fs::path& path, std::uint64_t byteOffset) noexcept {
+    if (assetManager == nullptr) {
+        return false;
+    }
+    
+    using namespace simdjson;
+
+    const auto filenameString = path.string();
+
+    AAsset* file = AAssetManager_open(assetManager, filenameString.c_str(), AASSET_MODE_BUFFER);
```
spnda, line 2230: Optional, but probably better is to (1) use `auto*` here and (2) wrap it with a unique_ptr and a custom deleter:
```cpp
auto assetDeleter = [](AAsset* file) { AAsset_close(file); }
std::unique_ptr<AAsset, decltype(assetDeleter)> file(AAssetManager_open(...), assetDeleter);
```
This guarantees that the file gets closed no matter if anything should throw an exception.

src/fastgltf_parser.hpp
```diff
@@ -308,14 +313,29 @@ namespace fastgltf {
...
         /**
          * Returns the size, in bytes,
          * @return
          */
         [[nodiscard]] inline std::size_t getBufferSize() const noexcept;
     };
 
+    #if defined(__ANDROID__)
+    class AndroidGltfDataBuffer : public GltfDataBuffer {
+        AAssetManager* assetManager = nullptr;
```
spnda, line 326: This assignment should still be gone, as its redundant. The constructor always initializes assetManager to something.

spnda (approved):

PR #15, opened 2023-04-09
Load external glTF buffers from Android assets

DethRaid:
This PR adds support for loading external buffers from within an APK

The method `fastgltf::glTF::loadFileFromUri` will try to load the file from the APK. If that fails, it will try to load the file from the filesystem as before

I've also removed a check for if the glTF's parent path is a filesystem directory. It is not a filesystem directory if it's in an APK

### DIFF
--- a/src/fastgltf.cpp
+++ b/src/fastgltf.cpp
@@ -456,7 +456,7 @@ std::pair<fg::Error, fg::DataSource> fg::glTF::loadFileFromApk(URI& uri) const n
         auto info = data->config.mapCallback(static_cast<std::uint64_t>(length), data->config.userPointer);
         if (info.mappedMemory != nullptr) {
             const sources::CustomBuffer customBufferSource = { info.customId, MimeType::None };
-            file.read(reinterpret_cast<char*>(info.mappedMemory), length);
+            AAsset_read(file.get(), info.mappedMemory, length);
             if (data->config.unmapCallback != nullptr) {
                 data->config.unmapCallback(&info, data->config.userPointer);
             }
--- a/src/fastgltf.cpp
+++ b/src/fastgltf.cpp
@@ -404,7 +404,7 @@ std::pair<fg::Error, fg::DataSource> fg::glTF::loadFileFromUri(URI& uri) const n
 
 #if defined(__ANDROID__)
     // Try to load external buffers from the APK. If they're not there, fall through to the file case
-    const auto android_result = loadFileFromApk(uri);
+    const auto android_result = loadFileFromApk(path);
     if(android_result.first == Error::None) {
         return android_result;
     }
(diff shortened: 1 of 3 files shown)


src/fastgltf.cpp
```diff
@@ -2394,10 +2450,12 @@ std::size_t fg::GltfDataBuffer::getBufferSize() const noexcept {
 
 #pragma region AndroidGltfDataBuffer
 #if defined(__ANDROID__)
-fg::AndroidGltfDataBuffer::AndroidGltfDataBuffer(AAssetManager* assetManager) noexcept : assetManager{assetManager} {}
+fg::AndroidGltfDataBuffer::AndroidGltfDataBuffer(AAssetManager* assetManager) noexcept {
+    androidAssetManager = assetManager;
```
spnda, line 2454: I think this is wrong to handle it like this. I could still open a glTF I loaded manually through a GltfDataBuffer and then the global would be nullptr. I think you should add a new global function that sets the global.

src/fastgltf.cpp
```diff
@@ -419,6 +439,42 @@ std::pair<fg::Error, fg::DataSource> fg::glTF::loadFileFromUri(URI& uri) const n
     return std::make_pair(Error::None, std::move(vectorSource));
 }
 
+#if defined(__ANDROID__)
+std::pair<fg::Error, fg::DataSource> fg::glTF::loadFileFromApk(const std::filesystem::path& filepath) const noexcept {
+    const auto fileString = filepath.string();
+    auto assetDeleter = [](AAsset* file) { AAsset_close(file); };
+    auto file = std::unique_ptr<AAsset, decltype(assetDeleter)>(AAssetManager_open(androidAssetManager, fileString.c_str(), AASSET_MODE_BUFFER), assetDeleter);
```
spnda, line 446: Perhaps these two lines could become a macro or something as they're duplicated from `loadFromAndroidAsset` and are 100% identical, and long.

src/fastgltf.cpp
```diff
@@ -390,6 +401,15 @@ std::pair<fg::Error, fg::DataSource> fg::glTF::decodeDataUri(URI& uri) const noe
 std::pair<fg::Error, fg::DataSource> fg::glTF::loadFileFromUri(URI& uri) const noexcept {
     auto path = directory / uri.path();
     std::error_code error;
+
+#if defined(__ANDROID__)
+    // Try to load external buffers from the APK. If they're not there, fall through to the file case
+    const auto android_result = loadFileFromApk(path);
+    if(android_result.first == Error::None) {
+        return android_result;
+    }
```
spnda, line 410: `loadFileFromApk` does not check if `androidAssetManager` is `nullptr` or not. Perhaps this should be handled here, and skip the APK load when the manager is null.

spnda (changes requested):

spnda:
I rebased & fixed up the PR and committed it: 5253adb

Thank you.

PR #20, opened 2023-04-26
Utilities to extract data from accessors

forenoonwatch:
This PR adds fastgltf_tools.hpp, a header containing several utility functions such as `getAccessorElement`, `iterateAccessor`, and `fillTargetFromAccessor`, the `ElementTraits<T>` struct for matching metadata about the desired type, and the `DefaultBufferDataAdapter`, a minimal default implementation of the `BufferDataAdapter` pattern which allows the accessor indexing functions to resolve used buffer memory.

Currently this PR lacks support for sparse accessors and compressed buffer views.

### DIFF
--- a/src/fastgltf_tools.hpp
+++ b/src/fastgltf_tools.hpp
@@ -143,33 +143,54 @@ ElementType getAccessorElementAt(ComponentType componentType, const std::byte* b
 
 }
 
-template <typename ElementType>
-ElementType getAccessorElement(const Asset& asset, const Accessor& accessor, const void* bufferData, size_t index) {
-	auto* bytes = reinterpret_cast<const std::byte*>(bufferData);
+struct DefaultBufferDataAdapter {
+	const std::byte* operator()(const Buffer& buffer) const {
+		const std::byte* result = nullptr;
 
+		std::visit([&](auto& arg) {
+			using SourceType = std::decay_t<decltype(arg)>;
+
+			if constexpr (std::is_same_v<SourceType, sources::Vector>) {
+				result = reinterpret_cast<const std::byte*>(arg.bytes.data());
+			}
+			else if constexpr (std::is_same_v<SourceType, sources::ByteView>) {
+				result = arg.bytes.data();
+			}
+		}, buffer.data);
+
+		return result;
+	}
+};
+
+template <typename ElementType, typename BufferDataAdapter = DefaultBufferDataAdapter>
+ElementType getAccessorElement(const Asset& asset, const Accessor& accessor, size_t index,
+		const BufferDataAdapter& adapter = {}) {
 	// FIXME: Assuming that I have a buffer view index, but what do I do if I don't?
 	auto& view = asset.bufferViews[*accessor.bufferViewIndex];
 	auto stride = view.byteStride ? *view.byteStride : getElementByteSize(accessor.type, accessor.componentType);
 
+	auto* bytes = adapter(asset.buffers[view.bufferIndex]);
 	bytes += view.byteOffset + accessor.byteOffset;
 
 	return internal::getAccessorElementAt<ElementType>(accessor.componentType, bytes + index * stride);
 }
 
-template <typename ElementType, typename Functor>
-void iterateAccessor(const Asset& asset, const Accessor& accessor, const void* bufferData, Functor&& func) {
+template <typename ElementType, typename Functor, typename BufferDataAdapter = DefaultBufferDataAdapter>
+void iterateAccessor(const Asset& asset, const Accessor& accessor, Functor&& func,
+		const BufferDataAdapter& adapter = {}) {
 	if (accessor.type != ElementTraits<ElementType>::type) {
 		return;
 	}
 
 	for (std::size_t i = 0; i < accessor.count; ++i) {
-		func(getAccessorElement<ElementType>(asset, accessor, bufferData));
+		func(getAccessorElement<ElementType>(asset, accessor, adapter));
 	}
 }
 
-template <typename ElementType, std::size_t TargetStride = sizeof(ElementType)>
-void fillTargetFromAccessor(const Asset& asset, const Accessor& accessor, const void* sourceBufferData,
-		void* dest) {
+template <typename ElementType, std::size_t TargetStride = sizeof(ElementType),
+		 typename BufferDataAdapter = DefaultBufferDataAdapter>
+void fillTargetFromAccessor(const Asset& asset, const Accessor& accessor, void* dest,
+		const BufferDataAdapter& adapter = {}) {
 	if (accessor.type != ElementTraits<ElementType>::type) {
 		return;
 	}
@@ -181,12 +202,12 @@ void fillTargetFromAccessor(const Asset& asset, const Accessor& accessor, const 
 
 	auto elemSize = getElementByteSize(accessor.type, accessor.componentType);
 
-	auto* srcBytes = reinterpret_cast<const std::byte*>(sourceBufferData) + view.byteOffset + accessor.byteOffset;
+	auto* srcBytes = adapter(asset.buffers[view.bufferIndex]) + view.byteOffset + accessor.byteOffset;
 	auto* dstBytes = reinterpret_cast<std::byte*>(dest);
 
 	if constexpr (std::is_trivially_copyable_v<ElementType>) {
 		if (srcStride == elemSize && srcStride == TargetStride) {
-			std::memcpy(dest, sourceBufferData, elemSize * accessor.count);
+			std::memcpy(dest, srcBytes, elemSize * accessor.count);
 		}
 		else {
 			for (std::size_t i = 0; i < accessor.count; ++i) {
--- a/src/fastgltf_tools.hpp
+++ b/src/fastgltf_tools.hpp
@@ -1,0 +1,206 @@
+/*
+ * Copyright (C) 2022 - 2023 spnda
+ * This file is part of fastgltf <https://github.com/spnda/fastgltf>.
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following
+ * conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+ * OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#pragma once
+
+#include "fastgltf_types.hpp"
+
+namespace fastgltf {
+
+template <typename>
+struct ElementTraits;
+
+template<>
+struct ElementTraits<std::int8_t> {
+	using element_type = std::int8_t;
+	using component_type = std::int8_t;
+	static constexpr auto type = AccessorType::Scalar;
+	static constexpr auto componentType = ComponentType::Byte;
+};
+
+template<>
+struct ElementTraits<std::uint8_t> {
+	using element_type = std::uint8_t;
+	using component_type = std::uint8_t;
+	static constexpr auto type = AccessorType::Scalar;
+	static constexpr auto componentType = ComponentType::UnsignedByte;
+};
+
+template<>
+struct ElementTraits<std::int16_t> {
+	using element_type = std::int16_t;
+	using component_type = std::int16_t;
+	static constexpr auto type = AccessorType::Scalar;
+	static constexpr auto componentType = ComponentType::Short;
+};
+
+template<>
+struct ElementTraits<std::uint16_t> {
+	using element_type = std::uint16_t;
+	using component_type = std::uint16_t;
+	static constexpr auto type = AccessorType::Scalar;
+	static constexpr auto componentType = ComponentType::UnsignedShort;
+};
+
+template<>
+struct ElementTraits<std::int32_t> {
+	using element_type = std::int32_t;
+	using component_type = std::int32_t;
+	static constexpr auto type = AccessorType::Scalar;
+	static constexpr auto componentType = ComponentType::Int;
+};
+
+template<>
+struct ElementTraits<std::uint32_t> {
+	using element_type = std::uint32_t;
+	using component_type = std::uint32_t;
+	static constexpr auto type = AccessorType::Scalar;
+	static constexpr auto componentType = ComponentType::UnsignedInt;
+};
+
+template<>
+struct ElementTraits<float> {
+	using element_type = float;
+	using component_type = float;
+	static constexpr auto type = AccessorType::Scalar;
+	static constexpr auto componentType = ComponentType::Float;
+};
+
+template<>
+struct ElementTraits<double> {
+	using element_type = double;
+	using component_type = double;
+	static constexpr auto type = AccessorType::Scalar;
+	static constexpr auto componentType = ComponentType::Double;
+};
+
+namespace internal {
+
+template <typename SourceType, typename DestType, std::size_t Index>
+constexpr DestType convertComponent(const std::byte* bytes) {
+	return static_cast<DestType>(reinterpret_cast<const SourceType*>(bytes)[Index]);
+}
+
+template <typename ElementType, typename SourceType, std::size_t... I>
+constexpr ElementType convertAccessorElement(const std::byte* bytes, std::index_sequence<I...>) {
+	using DestType = typename ElementTraits<ElementType>::component_type;
+
+	if constexpr (std::is_aggregate_v<ElementType>) {
+		return {convertComponent<SourceType, DestType, I>(bytes)...};
+	}
+	else {
+		return ElementType{convertComponent<SourceType, DestType, I>(bytes)...};
+	}
+}
+
+template <typename ElementType,
+		typename Seq = std::make_index_sequence<getNumComponents(ElementTraits<ElementType>::type)>>
+ElementType getAccessorElementAt(ComponentType componentType, const std::byte* bytes) {
+	switch (componentType) {
+		// This is undefined behavior if component type is invalid
+		default:
+		case ComponentType::Byte:
+			return convertAccessorElement<ElementType, std::int8_t>(bytes, Seq{});
+		case ComponentType::UnsignedByte:
+			return convertAccessorElement<ElementType, std::uint8_t>(bytes, Seq{});
+		case ComponentType::Short:
+			return convertAccessorElement<ElementType, std::int16_t>(bytes, Seq{});
+		case ComponentType::UnsignedShort:
+			return convertAccessorElement<ElementType, std::uint16_t>(bytes, Seq{});
+		case ComponentType::Int:
+			return convertAccessorElement<ElementType, std::int32_t>(bytes, Seq{});
+		case ComponentType::UnsignedInt:
+			return convertAccessorElement<ElementType, std::uint32_t>(bytes, Seq{});
+		case ComponentType::Float:
+			return convertAccessorElement<ElementType, float>(bytes, Seq{});
+		case ComponentType::Double:
+			return convertAccessorElement<ElementType, double>(bytes, Seq{});
+	}
+}
+
+}
+
+template <typename ElementType>
+ElementType getAccessorElement(const Asset& asset, const Accessor& accessor, const void* bufferData, size_t index) {
+	auto* bytes = reinterpret_cast<const std::byte*>(bufferData);
+
+	// FIXME: Assuming that I have a buffer view index, but what do I do if I don't?
+	auto& view = asset.bufferViews[*accessor.bufferViewIndex];
+	auto stride = view.byteStride ? *view.byteStride : getElementByteSize(accessor.type, accessor.componentType);
+
+	bytes += view.byteOffset + accessor.byteOffset;
+
+	return internal::getAccessorElementAt<ElementType>(accessor.componentType, bytes + index * stride);
+}
+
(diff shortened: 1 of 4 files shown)


src/fastgltf_types.hpp
```diff
@@ -260,7 +260,7 @@ namespace fastgltf {
     }
 
     constexpr std::uint32_t getElementByteSize(AccessorType type, ComponentType componentType) noexcept {
-        return getNumComponents(type) * (getComponentBitSize(componentType) / 8);
+        return (getNumComponents(type) * getComponentBitSize(componentType) + 7u) / 8u;
```
spnda, line 263: What's with this change?

tests/accessor_tests.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include <glm/gtx/matrix_decompose.hpp>
+
+#include "base64_decode.hpp"
+#include "fastgltf_parser.hpp"
+#include "fastgltf_types.hpp"
+#include "fastgltf_tools.hpp"
+#include "gltf_path.hpp"
+
+template<class... Ts> struct overloaded : Ts... { using Ts::operator()...; };
+template<class... Ts> overloaded(Ts...) -> overloaded<Ts...>;
```
spnda, line 20: As they're now doubled in gl_viewer and here, I think these should be moved into fastgltf_util.hpp as a convenience tool for other developers.

tests/accessor_tests.cpp
```diff
@@ -0,0 +1,163 @@
...
+    REQUIRE(parser.getError() == fastgltf::Error::None);
+    REQUIRE(model->parse(fastgltf::Category::Buffers | fastgltf::Category::BufferViews
+			| fastgltf::Category::Accessors) == fastgltf::Error::None);
+    REQUIRE(model->validate() == fastgltf::Error::None);
+
+    auto asset = model->getParsedAsset();
+    REQUIRE(asset->accessors.size() == 15);
+    auto& accessors = asset->accessors;
+
+    {
```
spnda, line 62: Perhaps you could make these two things a `SECTION` because they test two different things.

tests/accessor_tests.cpp
```diff
@@ -0,0 +1,163 @@
...
+		REQUIRE(firstAccessor.bufferViewIndex.has_value());
+		auto& view = asset->bufferViews[*firstAccessor.bufferViewIndex];
+
+		auto* bufferData = getBufferData(asset->buffers[view.bufferIndex]);
+		REQUIRE(bufferData != nullptr);
+
+		auto* checkData = reinterpret_cast<const unsigned short*>(bufferData + view.byteOffset
+				+ firstAccessor.byteOffset);
+
+		REQUIRE(*checkData == fastgltf::getAccessorElement<unsigned short>(*asset, firstAccessor, 0));
```
spnda, line 76: Please use `std::uint16_t` instead of `unsigned short`.

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,303 @@
...
+		for (std::size_t i = 0; i < accessor.count; ++i) {
+			func(internal::getAccessorElementAt<ElementType>(accessor.componentType, bytes + i * stride));
+		}
+	}
+
+}
+
+template <typename ElementType, std::size_t TargetStride = sizeof(ElementType),
+		 typename BufferDataAdapter = DefaultBufferDataAdapter>
+void fillTargetFromAccessor(const Asset& asset, const Accessor& accessor, void* dest,
```
spnda, line 243: I feel like `copyFromAccessor` is a much better general name for this function.

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,303 @@
...
+		auto stride = view.byteStride ? *view.byteStride : getElementByteSize(accessor.type, accessor.componentType);
+
+		auto* bytes = adapter(asset.buffers[view.bufferIndex]);
+		bytes += view.byteOffset + accessor.byteOffset;
+
+		for (std::size_t i = 0; i < accessor.count; ++i) {
+			func(internal::getAccessorElementAt<ElementType>(accessor.componentType, bytes + i * stride));
+		}
+	}
+
```
spnda, line 238: empty line

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,303 @@
...
+			auto idx = internal::getAccessorElementAt<std::uint32_t>(accessor.sparse->indexComponentType,
+					indicesBytes + indexStride * i);
+			auto* pDest = reinterpret_cast<ElementType*>(dstBytes + TargetStride * i);
+
+			*pDest = internal::getAccessorElementAt<ElementType>(accessor.componentType,
+					valuesBytes + srcStride * idx);
+		}
+	}
+	else {
+		// FIXME: Assuming that I have a buffer view index, but what do I do if I don't?
```
spnda, line 274: Here's what you do:
> 5.1.1. accessor.bufferView
The index of the buffer view. When undefined, the accessor MUST be initialized with zeros; sparse property or extensions MAY override zeros with actual values.

Same applies to the `getAccessorElement` function which also doesn't check if a value is given.

spnda (changes requested):
I have a few general things:
- Please adapt the `else` style to the rest of the repo, as well as the naming (currently snake_case) of the template specialisation constants.
- Probably should add some static_asserts that the required ElementTraits all exist, and that certain template types meet some criteria (could use concepts with a C++20 preprocessor guard).

src/fastgltf_types.hpp
forenoonwatch: Probably not strictly necessary since all current glTF types are powers of 2, but this is the correct way to pack any sized components into a certain number of bytes. I felt the need to correct this since it would otherwise be weird to bother working in bit sizes for the components.

tests/accessor_tests.cpp
forenoonwatch: Seems like a good idea, although they would probably need a more descriptive name. In my own engine I call these "VisitorBase" since their job is to serve as a visitor base class.

src/fastgltf_tools.hpp
forenoonwatch: Resolved in [bc67bc4](https://github.com/spnda/fastgltf/pull/20/commits/bc67bc47bf712fc37adb1fbdd9f92ddeb943cab4)

src/fastgltf_tools.hpp
forenoonwatch: Resolved in [bc67bc4](https://github.com/spnda/fastgltf/pull/20/commits/bc67bc47bf712fc37adb1fbdd9f92ddeb943cab4)

tests/accessor_tests.cpp
forenoonwatch: Resolved in [fd17efd](https://github.com/spnda/fastgltf/pull/20/commits/fd17efd965946090d4400e324b488e40861ae31a)

tests/accessor_tests.cpp
forenoonwatch: Resolved in [fd17efd](https://github.com/spnda/fastgltf/pull/20/commits/fd17efd965946090d4400e324b488e40861ae31a)

src/fastgltf_tools.hpp
forenoonwatch: Resolved in [472b1db](https://github.com/spnda/fastgltf/pull/20/commits/472b1dbdffc920d8731b29e0cb1e87d62126b1a1)

tests/accessor_tests.cpp
spnda: Or just simply "visitor". It would ultimately be fastgltf::visitor which I think would be perfectly fine.

spnda (changes requested):
- I'd like to see more comments around the top level functions and in some complex situations for the future.
- In Release mode on MSVC the "getAccessorElement" test fails sometimes (!) with sparse accessors. With a debugger I noticed the values are *definitely* wrong but I couldn't tell you why.

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,376 @@
...
+constexpr ElementType convertAccessorElement(const std::byte* bytes, std::index_sequence<I...>) {
+	using DestType = typename ElementTraits<ElementType>::component_type;
+	static_assert(std::is_arithmetic_v<DestType>, "Accessor traits must provide a valid component ttype");
+
+	if constexpr (std::is_aggregate_v<ElementType>) {
+		return {convertComponent<SourceType, DestType, I>(bytes)...};
+	} else {
+		return ElementType{convertComponent<SourceType, DestType, I>(bytes)...};
+	}
+}
```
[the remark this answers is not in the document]
forenoonwatch: Yeah that's exactly it, the parameter is just to be able to deduce the underlying sequence as a variadic so that I can fold on it.

tests/accessor_tests.cpp
forenoonwatch: Added fastgltf::visitor in [9a35575](https://github.com/spnda/fastgltf/pull/20/commits/9a35575e79810ed5ef622f8f487d83f5f9fab50b).

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,376 @@
...
+
+		std::visit([&](auto& arg) {
+			using SourceType = std::decay_t<decltype(arg)>;
+
+			if constexpr (std::is_same_v<SourceType, sources::Vector>) {
+				result = reinterpret_cast<const std::byte*>(arg.bytes.data());
+			} else if constexpr (std::is_same_v<SourceType, sources::ByteView>) {
+				result = arg.bytes.data();
+			}
+		}, buffer.data);
```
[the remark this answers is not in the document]
forenoonwatch: Changed in [64b40b7](https://github.com/spnda/fastgltf/pull/20/commits/64b40b79c8211fd3b9f024bc44ced66911a3feaf)

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,376 @@
...
+		// "The index of the bufferView with sparse values. The referenced buffer view MUST NOT
+		// have its target or byteStride properties defined."
+		auto valueStride = getElementByteSize(accessor.type, accessor.componentType);
+
+		auto idx = internal::getAccessorElementAt<std::uint32_t>(accessor.sparse->indexComponentType,
+				indicesBytes + indexStride * index);
+
+		return internal::getAccessorElementAt<ElementType>(accessor.componentType,
+				valuesBytes + valueStride * idx);
+	} else {
```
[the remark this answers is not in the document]
forenoonwatch: Fixed in [e973b07](https://github.com/spnda/fastgltf/pull/20/commits/e973b071307d98acb319139aa1559b8afee81910)

tests/accessor_tests.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include <glm/glm.hpp>
+#include <glm/gtc/type_ptr.hpp>
+#include <glm/gtx/quaternion.hpp>
+#include <glm/gtx/matrix_decompose.hpp>
+
+#include "base64_decode.hpp"
+#include "fastgltf_parser.hpp"
+#include "fastgltf_types.hpp"
+#include "fastgltf_tools.hpp"
+#include "gltf_path.hpp"
```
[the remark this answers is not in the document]
forenoonwatch: Fixed in [f2b62bd](https://github.com/spnda/fastgltf/pull/20/commits/f2b62bd0a056aae4c2808f9b7c45b4e1dcae7839)

forenoonwatch:
> * In Release mode on MSVC the "getAccessorElement" test fails sometimes (!) with sparse accessors. With a debugger I noticed the values are _definitely_ wrong but I couldn't tell you why.
> 
> > REQUIRE(checkValues[i] == fastgltf::getAccessorElementglm::vec3(*asset, secondAccessor, i))
> > with expansion:
> > {?} == {?}

I've been able to replicate this in my MinGW release (intermittently). Trying to see if I can actually figure out what's causing it.

tests/accessor_tests.cpp
```diff
@@ -0,0 +1,177 @@
...
+			REQUIRE(checkValues[i] == fastgltf::getAccessorElement<glm::vec3>(*asset, secondAccessor, i));
+		}
+	}
+
+	SECTION("iterateAccessor") {
+		auto dstCopy = std::make_unique<glm::vec3[]>(secondAccessor.count);
+		std::size_t i = 0;
+
+		fastgltf::iterateAccessor<glm::vec3>(*asset, secondAccessor, [&](auto&& v3) {
+			dstCopy[i++] = std::move(v3);
```
spnda, line 166: Use `std::forward`.

tests/accessor_tests.cpp
```diff
@@ -0,0 +1,177 @@
...
+		SECTION("getAccessorElement<glm::vec3>") {
+			REQUIRE(*checkData == fastgltf::getAccessorElement<glm::vec3>(*asset, secondAccessor, 0));
+		}
+
+		SECTION("iterateAccessor") {
+			auto dstCopy = std::make_unique<glm::vec3[]>(secondAccessor.count);
+			std::size_t i = 0;
+
+			fastgltf::iterateAccessor<glm::vec3>(*asset, secondAccessor, [&](auto&& v3) {
+				dstCopy[i++] = std::move(v3);
```
spnda, line 89: Use `std::forward`.

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,467 @@
...
+
+	if (accessor.sparse && accessor.sparse->count > 0) {
+		std::size_t index = 0;
+		iterateAccessor<ElementType>(asset, accessor, [&](auto&& value) {
+			auto* pDest = reinterpret_cast<ElementType*>(dstBytes + TargetStride * index);
+			*pDest = std::move(value);
+			++index;
+		}, adapter);
+	} else {
+		auto elemSize = getElementByteSize(accessor.type, accessor.componentType);
```
spnda, line 415: This should probably not be nested. I think you should return after the iterateAccessor and then remove the else, like you do a little further down.

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,467 @@
...
+				}
+			} else if (accessor.bufferViewIndex) {
+				func(internal::getAccessorElementAt<ElementType>(accessor.componentType,
+						srcBytes + srcStride * i));
+			} else {
+				func(ElementType{});
+			}
+		}
+	} else {
+		// 5.1.1. accessor.bufferView
```
spnda, line 366: This should probably also be a `return` so that the else can be removed.

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,467 @@
...
+		case ComponentType::UnsignedInt:
+			return findSparseIndex<std::uint32_t>(bytes, indexCount, desiredIndex, resultIndex);
+		default:
+			break;
+	}
+
+	return false;
+}
+
+}
```
spnda, line 210: `} // namespace internal`

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,467 @@
...
+
+template <typename ElementType,
+		typename Seq = std::make_index_sequence<getNumComponents(ElementTraits<ElementType>::type)>>
+#if FASTGLTF_HAS_CONCEPTS
+requires Element<ElementType>
+#endif
+ElementType getAccessorElementAt(ComponentType componentType, const std::byte* bytes) {
+	switch (componentType) {
+		// This is undefined behavior if component type is invalid
+		default:
```
spnda, line 142: I seriously think we shouldn't have a `default` branch here. If I were to add a new component type, the compiler would warn me about a missing branch here. With the `default` branch, we gain nothing but catching UB (casting to a enum class with a non-existent value) that isn't even caused by us. See this:
> The static_cast operator can explicitly convert an integral value to an enumeration type. If the value of the integral type does not fall within the range of enumeration values, the resulting enumeration value is undefined.

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,467 @@
...
+		case ComponentType::Short:
+			return findSparseIndex<std::int16_t>(bytes, indexCount, desiredIndex, resultIndex);
+		case ComponentType::UnsignedShort:
+			return findSparseIndex<std::uint16_t>(bytes, indexCount, desiredIndex, resultIndex);
+		case ComponentType::Int:
+			return findSparseIndex<std::int32_t>(bytes, indexCount, desiredIndex, resultIndex);
+		case ComponentType::UnsignedInt:
+			return findSparseIndex<std::uint32_t>(bytes, indexCount, desiredIndex, resultIndex);
+		default:
+			break;
```
spnda, line 204: Again, please don't use a `default` branch, but instead cover the `Float` and `Double` cases with a simple `return false;`

spnda (changes requested):

src/fastgltf_tools.hpp
forenoonwatch: Fixed in [df17edc](https://github.com/spnda/fastgltf/pull/20/commits/df17edc742f1893ece80c0cc2467a9b68b7ec880)

src/fastgltf_tools.hpp
forenoonwatch: Fixed in [df17edc](https://github.com/spnda/fastgltf/pull/20/commits/df17edc742f1893ece80c0cc2467a9b68b7ec880)

src/fastgltf_tools.hpp
forenoonwatch: Fixed in [df17edc](https://github.com/spnda/fastgltf/pull/20/commits/df17edc742f1893ece80c0cc2467a9b68b7ec880)

src/fastgltf_tools.hpp
forenoonwatch: Fixed in [df17edc](https://github.com/spnda/fastgltf/pull/20/commits/df17edc742f1893ece80c0cc2467a9b68b7ec880)

tests/accessor_tests.cpp
forenoonwatch: Fixed in [df17edc](https://github.com/spnda/fastgltf/pull/20/commits/df17edc742f1893ece80c0cc2467a9b68b7ec880)

tests/accessor_tests.cpp
forenoonwatch: Fixed in [df17edc](https://github.com/spnda/fastgltf/pull/20/commits/df17edc742f1893ece80c0cc2467a9b68b7ec880)

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,486 @@
...
+	if (!accessor.bufferViewIndex) {
+		if constexpr (std::is_trivially_copyable_v<ElementType>) {
+			if (TargetStride == elemSize) {
+				std::memset(dest, 0, elemSize * accessor.count);
+			} else {
+				for (std::size_t i = 0; i < accessor.count; ++i) {
+					std::memset(dstBytes + i * TargetStride, 0, elemSize);
+				}
+			}
+		} else {
```
spnda, line 449: `std::is_trivially_copyable_v` does not guarantee that a "default constructed" value is zero'd. It only guarantees that you can `std::memcpy` between a byte representation or other objects of the same type. I think the sensible option would be to check `std::is_trivially_constructible_v<ElementType, std::uint32_t>` (or std::byte?) to let them default construct with `0`: `ElementType { 0 }`. Though then you could also just use the branch below.

Or am I missing something here?

forenoonwatch: This is a valid concern, however the highlighted code here is actually the correct part. using `std::memset` with trivially_copyable_v basically pretends that the source data is a bunch of zeroes and we are copying these zero bytes from them.

The problem arises in the other branch where we make the assumption that `ElementType{}` is the same as zero-initializing, but ultimately I think this is a safe assumption to make, and the alternate solutions might involve a significant amount of template trickery and an increased number of assumptions about the implementation of ElementType.
For example, ElementType{0} on a non-scalar type assumes that it A) has a single-arg constructor B) would correctly zero-fill if its first argument is 0. Another assumption we could make to fix that is to assume a ComponentCount-arg constructor, and come up with some templating to zero-fill the constructor with n zeroes, but that still makes the assumption that such a constructor exists and produces the correct result.

forenoonwatch: After thinking about it some more, one possible solution is to simply request the user provide a default value, a la `std::vector.resize(size_type count, const value_type& value)`. This has the benefit of being more extensible by the user and makes the least assumptions about present constructors, but requires the user to be aware that some of their math types may need to have their zero-value be explicitly specified. This may also result in code duplication, since a 1-arg and 2-arg version would have to be implemented separately, as with vector resize, to minimize assumptions about copyability and constructability.

In all, I think this is a bit overkill considering the types specified are all numeric types, and it should be a safe assumption to make in general that the default constructor of a numeric type is the equivalent of zero in that number space.

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,478 @@
...
+	return static_cast<DestType>(reinterpret_cast<const SourceType*>(bytes)[Index]);
+}
+
+template <typename ElementType, typename SourceType, std::size_t... I>
+#if FASTGLTF_HAS_CONCEPTS
+requires Element<ElementType>
+#endif
+constexpr ElementType convertAccessorElement(const std::byte* bytes, std::index_sequence<I...>) {
+	using DestType = typename ElementTraits<ElementType>::component_type;
+	static_assert(std::is_arithmetic_v<DestType>, "Accessor traits must provide a valid component ttype");
```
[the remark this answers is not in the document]
forenoonwatch: Fixed in [a98bca0](https://github.com/spnda/fastgltf/pull/20/commits/a98bca0a30247a9dcbf0eea2a74c932e7750bb57)

src/fastgltf_tools.hpp
```diff
@@ -0,0 +1,478 @@
...
+	static constexpr auto type = ComponentType::Float;
+};
+
+template<>
+struct ComponentTypeConverter<double> {
+	static constexpr auto type = ComponentType::Double;
+};
+
+template <typename ElementType, typename ComponentType, AccessorType EnumAccessorType>
+struct ElementTraitsBase {
```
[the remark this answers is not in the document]
forenoonwatch: Seems like a pretty good idea, added in [a98bca0](https://github.com/spnda/fastgltf/pull/20/commits/a98bca0a30247a9dcbf0eea2a74c932e7750bb57).

tests/accessor_tests.cpp
```diff
@@ -0,0 +1,177 @@
...
+    REQUIRE(asset->accessors[1].sparse.has_value());
+    auto& sparse = asset->accessors[1].sparse.value();
+    REQUIRE(sparse.count == 3);
+    REQUIRE(sparse.bufferViewIndices == 2);
+    REQUIRE(sparse.byteOffsetIndices == 0);
+    REQUIRE(sparse.bufferViewValues == 3);
+    REQUIRE(sparse.byteOffsetValues == 0);
+    REQUIRE(sparse.indexComponentType == fastgltf::ComponentType::UnsignedShort);
+
+
```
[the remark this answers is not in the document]
forenoonwatch: Fixed in [a98bca0](https://github.com/spnda/fastgltf/pull/20/commits/a98bca0a30247a9dcbf0eea2a74c932e7750bb57)

src/fastgltf_tools.hpp
spnda: Yeah I guess it's better the way it is currently. I think this is fine as-is, and if anybody complains about needing some specific use-case we can see later. I guess we can leave it, and perhaps add some checks in the future to ensure that the types are numeric or can be constructed properly from "nothing".

PR #21, opened 2023-05-09
Fix URI moves for heap-allocated strings

Eearslya:
Funnily enough, the last fix which added `readjustViews` fixed the problem with stack-allocated strings, but broke heap-allocated strings. Since the move constructor is no longer default, the string views no longer get moved automatically, and now we have to do it.


spnda (changes requested):
`std::string_view` does not have any move constructors. Simply copying here would be sufficient. Also while you're here you could implement the thing you mentioned with `*this = other;`?

Eearslya:
Done and done.

spnda (approved):

PR #30, opened 2023-09-05
Added support for EXT_mesh_gpu_instancing

DragonJoker:


include/fastgltf/parser.hpp
```diff
@@ -265,6 +268,7 @@ namespace fastgltf {
         constexpr std::string_view KHR_materials_unlit = "KHR_materials_unlit";
         constexpr std::string_view KHR_materials_volume = "KHR_materials_volume";
         constexpr std::string_view KHR_mesh_quantization = "KHR_mesh_quantization";
+        constexpr std::string_view EXT_mesh_gpu_instancing = "EXT_mesh_gpu_instancing";
```
spnda, line 271: why is this just put randomly in between the KHR extensions? these are sorted alphabetically.

include/fastgltf/parser.hpp
```diff
@@ -289,6 +293,7 @@ namespace fastgltf {
 		{ extensions::KHR_materials_unlit,                Extensions::KHR_materials_unlit },
 		{ extensions::KHR_materials_volume,               Extensions::KHR_materials_volume },
 		{ extensions::KHR_mesh_quantization,              Extensions::KHR_mesh_quantization },
+		{ extensions::EXT_mesh_gpu_instancing,            Extensions::EXT_mesh_gpu_instancing },
```
spnda, line 296: same as above

include/fastgltf/types.hpp
```diff
@@ -1309,6 +1309,11 @@ namespace fastgltf {
          */
         std::variant<TRS, TransformMatrix> transform;
 
+        /**
+         * Only ever non-empty when EXT_mesh_gpu_instancing is enabled and used by the asset.
+         */
+        pmr::SmallVector<std::pair<std::pmr::string, std::size_t>, 4> instancingAttributes;
```
spnda, line 1315: Would it make sense to add a function like the findAttribute function on primitives? Seems like they have the same signature and same basic concept.

spnda (changes requested):

include/fastgltf/parser.hpp
DragonJoker: Ah, wrong gues, then, I thought it was about what it was related to (mesh, in this case)

include/fastgltf/types.hpp
DragonJoker: Definitely !

DragonJoker: And done (I've also added some `noexcept`)

include/fastgltf/parser.hpp
DragonJoker: Fixed.

include/fastgltf/parser.hpp
DragonJoker: Fixed.

include/fastgltf/types.hpp
```diff
@@ -1309,8 +1309,29 @@ namespace fastgltf {
...
```
spnda, line 1315: I want extensions to be more or less invisible if the user is not using any of them, so having 4 counts of the std::pair *always* part of the Node struct seems weird and like a waste of memory to me when people are not using the extension. This is actually 128 bytes for all the pairs. (24 bytes for string, 8 bytes for size_t, each times four).

I'd want either one of these:
- Wrap the SmallVector in a `unique_ptr` so that this is only 8 bytes. I do this with all other extensions, too.
- Simply use a ordinary `pmr::vector` instead of the `SmallVector`. This is still 24 bytes, but would drastically reduce it from 152 bytes.

spnda (changes requested):

include/fastgltf/types.hpp
DragonJoker: I've replaced it with a pmr::vector.

PR #31, opened 2023-09-06
Added support for KHR_materials_pbrSpecularGlossiness.

DragonJoker:


include/fastgltf/parser.hpp
```diff
@@ -268,13 +273,21 @@ namespace fastgltf {
         constexpr std::string_view KHR_texture_basisu = "KHR_texture_basisu";
         constexpr std::string_view KHR_texture_transform = "KHR_texture_transform";
         constexpr std::string_view MSFT_texture_dds = "MSFT_texture_dds";
+#if FASTGLTF_ENABLE_DEPRECATED_EXT
+        constexpr std::string_view KHR_materials_pbrSpecularGlossiness = "KHR_materials_pbrSpecularGlossiness";
+#endif
```
DragonJoker, line 278: I haven't respected the alphabetical order here, to eventually group the deprecated extensions in the same area.
Tell me if you want me to change that.

spnda: yeah seems fine. Though if you wanna do it like that a empty line  between them would make sense imo.

DragonJoker: Like it is now ?

spnda: yup

include/fastgltf/parser.hpp
```diff
@@ -174,11 +174,16 @@ namespace fastgltf {
...
-	    KHR_materials_anisotropy = 1 << 18,
+        // See https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
+        KHR_materials_anisotropy = 1 << 18,
 
         // See https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/EXT_mesh_gpu_instancing/README.md
         EXT_mesh_gpu_instancing = 1 << 19,
+
+#if FASTGLTF_ENABLE_DEPRECATED_EXT
+        // See https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Archived/KHR_materials_pbrSpecularGlossiness/README.md
+        KHR_materials_pbrSpecularGlossiness = 1 << 20,
```
spnda, line 185: I guess it could also be useful to mark this [[deprecated]], which is allowed since C++17:
```cpp
KHR_materials_pbrSpecularGlossiness [[deprecated]] = 1 << 20
```

spnda:
I guess it would also make sense to add a Actions runner that would also run the tests with the define enabled. Right now, the tests never get built through CI for the extension.
If you're unsure how to do that, I can also take care of it after the PR is merged.

include/fastgltf/parser.hpp
DragonJoker: It's not a deprecated feature of the fastgltf code but of the glTF format.
I wonder why we would tell the user he is using deprecated code (since he is not)

spnda: I was thinking of using it as an indicator about the state of the glTF extension and not the actual code being deprecated. But sure, was only a thought anyway.

include/fastgltf/types.hpp
```diff
@@ -1502,6 +1502,19 @@ namespace fastgltf {
...
+#if FASTGLTF_ENABLE_DEPRECATED_EXT
+    /**
+     * Specular/Glossiness information from KHR_materials_pbrSpecularGlossiness.
+     */
+    struct MaterialSpecularGlossiness {
+        std::array<float, 4> diffuseFactor;
+        std::optional<TextureInfo> diffuseTexture;
+        std::array<float, 3> specularFactor;
+        float glossinessFactor;
+        std::optional<TextureInfo> specularGlossinessTexture;
```
spnda, line 1514: Replace std::optional with Optional. Its a type alias that checks wether a specialisation for fastgltf::OptionalWithFlagValue exists for T. When I do specialise it in the future for TextureInfo, I wouldn't have to change these lines.

spnda (changes requested):

include/fastgltf/types.hpp
DragonJoker: Done

spnda (approved):

PR #47, opened 2024-02-11
Fix: Inline deserializeComponent template specializations

hilloftheking:
This allows for multiple translation units to include tools.hpp
For some reason this already worked properly on GCC, but now it will work on MSVC.

### DIFF
--- a/include/fastgltf/tools.hpp
+++ b/include/fastgltf/tools.hpp
@@ -145,7 +145,7 @@ template<>
 #if FASTGLTF_CONSTEXPR_BITCAST
 constexpr
 #endif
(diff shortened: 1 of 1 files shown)


spnda (approved):

spnda:
After merging I noticed that perhaps some compilers will throw a warning about the function being marked as `constexpr inline` when `FASTGLTF_CONSTEXPR_BITCAST` is `1`. I can only find a relevant MR from clang-tidy which shows a warning in this case: https://reviews.llvm.org/D18914

If that ever comes up from someone using -Wall or -Werror i'll fix it. Thanks though.

PR #57, opened 2024-03-12
Fix: Bad chunk padding for GLB export

Cyphall:
This PR fixes two bugs:

1) The trailing padding for the BIN chunk was applied between `chunkType` and `chunkData` instead of after `chunkData` (terms from the [glTF Specs](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#chunks-overview))
2) The padding code was not adding the right amount of bytes (e.g. for a chunk size of 11, it was adding 3 bytes instead of 1).

I also refactored a few lines to have the same code structure of "write and pad" for the JSON and BIN chunks


spnda (approved):

spnda:
Thank you, really stupid mistake of having the padding in front of the chunk data...

PR #61, opened 2024-05-11
Support C++20 module with optional standard library module.

stripe2933:
This PR adds C++20 module feature with optional standard library module (can be enabled via `FASTGLTF_USE_STD_MODULE` macro).

- `fastgltf.cppm` module file exports symbols. Note that helper classes/functions from `base64.hpp`, `util.hpp` are not exported and therefore invisible to the end user.
- Do not include standard library headers if `FASTGLTF_USE_STD_MODULE` macro defined, and use std module instead. This can dramatically reduce the compile time.


module/fastgltf.cppm
```diff
@@ -0,0 +1,162 @@
...
+import std;
+import std.compat;
+
+extern "C++" {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Winclude-angled-in-module-purview"
+#include <fastgltf/core.hpp>
+#pragma clang diagnostic pop
+}
+#endif
```
spnda, line 20: What is the reason for these clang diagnostics? Looking at [this issue](https://github.com/llvm/llvm-project/issues/68615), it doesn't look like we should be including the header before the `export module` declaration.

module/fastgltf.cppm
```diff
@@ -0,0 +1,162 @@
...
+    using fastgltf::operator|;
+    using fastgltf::operator&;
+    using fastgltf::operator|=;
+    using fastgltf::operator&=;
+    using fastgltf::operator~;
+
+    using fastgltf::validate;
+    using fastgltf::getElementByteSize;
+    using fastgltf::getErrorName;
+    using fastgltf::getErrorMessage;
```
spnda, line 161: I don't like this having any indentation. Should move this one to the left or move everything else out by one, similar to the other headers.

module/fastgltf.cppm
```diff
@@ -0,0 +1,162 @@
...
+    using math::scale;
+    using math::rotate;
+    using math::transpose;
+    using math::fmat2x2;
+    using math::fmat3x3;
+    using math::fmat4x4;
+    using math::dmat2x2;
+    using math::dmat3x3;
+    using math::dmat4x4;
+    using math::decomposeTransformMatrix;
```
spnda, line 75: The order of these is not fully identical to what is in the header. Perhaps it'd be nicer to split this up into types, aliases, and then functions. Or at least have some space between the "sections".

spnda (changes requested):
How is `FASTGLTF_USE_STD_MODULE` expected to be defined? How does this work together with CMake?

Also, this doesn't include the changes from #49.

module/fastgltf.cppm
stripe2933: I think use `#include` in `extern "C++" { ... }` clause is okay to do. [This is from Clang documentation](https://clang.llvm.org/docs/StandardCPlusPlusModules.html#export-extern-c-style).

The reason I include the header file in export declaration region is: If `FASTGLTF_USE_STD_MODULE` macro defined, the definition of `std::`-prefixed things should be known before `<fastgltf/core.hpp>` header inclusion, however I can't place `import std;` inside the global module fragment, such like:

```c++
module;

import std; // Can't be here!
#include <fastgltf/core.hpp>

export module fastgltf;

// ...
```

Therefore, if `FASTGLTF_USE_STD_MODULE` macro not defined (and consequently not using `std` module), header inclusion should be inside the global module fragment. But otherwise, it have to be in export declaration regions, wrapped with `extern "C++" { ... }` clause.

stripe2933:
I created the [test repository](https://github.com/stripe2933/fastgltf-module-test) with latest Clang CI (runs on ubuntu and macOS). For now, it has dependency of [my another repository](https://github.com/stripe2933/CppStandardLibraryModule) for using `std` module. But CMake ≥ 3.30 will support it natively, so I'm looking forward that the CI time would be reduced.

module/fastgltf.cppm
stripe2933: Fixed.

CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ option(FASTGLTF_ENABLE_DEPRECATED_EXT "Enables support for deprecated extensions
 option(FASTGLTF_DISABLE_CUSTOM_MEMORY_POOL "Disables the memory allocation algorithm based on polymorphic resources" OFF)
 option(FASTGLTF_USE_64BIT_FLOAT "Default to 64-bit double precision floats for everything" OFF)
 option(FASTGLTF_COMPILE_AS_CPP20 "Have the library compile as C++20" OFF)
+option(FASTGLTF_INSTALL "Install fastgltf" OFF) # For CI.
```
spnda, line 21: Why add this option exactly? CMake install commands only configure a special installation target for the targets, so if you never run `cmake --install .` those commands won't ever run.

module/fastgltf.cppm
```diff
@@ -0,0 +1,166 @@
+module;
+
+#include <cassert>
```
spnda, line 3: Is this a bug with the core header not including `cassert` itself?

module/fastgltf.cppm
spnda: You missed `MappedGltfFile`, but otherwise this seems complete now, I think.

module/fastgltf.cppm
spnda: And I take it the `import std` is a C++23 feature? And using that has compile time benefits? Sorry, I'm quite unfamiliar with how modules work exactly. Though I guess this is fine.

spnda:
> `FASTGLTF_USE_STD_MODULE` can be defined as CMake `target_compile_definitions`command, like [this](https://github.com/stripe2933/fastgltf-module-test/blob/b7eeae6131b346287f8175f2b0e16c2016750d39/CMakeLists.txt#L42).

Would it perhaps make sense to have this auto-defined when using a supported compiler? Or could that break things?

> But CMake ≥ 3.30 will support it natively, so I'm looking forward that the CI time would be reduced.

I guess I'll just add CI runners for the modules later when that releases. Right now, I don't think I want to convoluted the CMake codebase for that.

module/fastgltf.cppm
stripe2933: core.hpp includes type.hpp, which uses assert macro. Since a module does not export macro, <cassert> have to be manually included in global module fragment.

module/fastgltf.cppm
stripe2933: > And I take it the import std is a C++23 feature?

`import std` is C++23 feature [(P2465R3)](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2465r3.pdf), but [major compiler vendors approved to support the feature in C++20](https://github.com/microsoft/STL/issues/3945) (I think it's because it is easier to adopt `std` module than header units).

> And using that has compile time benefits?

Yes, module compiled only once, therefore `import fastgltf;` doesn't introduce any compile time overhead. Also it is likely to user already compiled `std` module, so reusing it is reasonable.

module/fastgltf.cppm
stripe2933: Separated them by vector/quaternion/function, types/aliases/functions in order.

CMakeLists.txt
stripe2933: My mistake. Reverted.

module/fastgltf.cppm
stripe2933: Fixed.

module/fastgltf.cppm
spnda: But wouldn't that only apply to when we use `import std;` then, which wouldn't export the macro? `cassert` is included normally from `types.hpp`.

stripe2933: What you say is if FASTGLTF_USE_STD_MODULE macro not defined, cassert header is already included in types.hpp, therefore it doesn't need to be included in the global module fragment? Am I understanding right?

spnda: Yeah. I am currently writing some CMake stuff locally to support this addition out-of-the-box which I will push later onto your fork, and I have removed it locally and that doesn't cause any issues.

spnda:
I've tested my CMake changes locally and they seem to work just fine on Clang 18 and VS 17.9. It also compiles on GCC 14.1 but it takes forever to link. I've just added a warning for now about the `import std` stuff being used on pre-C++23.

stripe2933:
Well, CIs failed to build in my case. [stripe2933/fastgltf-module-test@69a28f5](https://github.com/stripe2933/fastgltf-module-test/tree/69a28f5)

When using [my `std` module enable strategy](https://github.com/stripe2933/CppStandardLibraryModule), it links `std` and `std.compat` targets to the all targets in the directories using `link_libraries` (which is independent to the destination target). It is conflicted with your `fastgltf::module` target, because:

- If `std` target declaration is [before](https://github.com/stripe2933/fastgltf-module-test/actions/runs/9058919976/job/24885513702) the `fastgltf::module` target declaration, `fastgltf::module` cannot use `std` target unless it explicitly linked (e.g. `target_link_libraries(fastgltf_module PUBLIC std std.compat)`). I think it is unintuitive.
- If `std` target declaration is [after](https://github.com/stripe2933/fastgltf-module-test/actions/runs/9058919976/job/24885513529) the `fastgltf::module` target declaration, CMake configure failed, because `fastgltf` target links `std` module but does not export the `std` target.
    ```
    CMake Error: install(EXPORT "fastgltf-targets" ...) includes target "fastgltf" which requires target "std" that is not in any export set.
    CMake Error: install(EXPORT "fastgltf-targets" ...) includes target "fastgltf" which requires target "std.compat" that is not in any export set.
    ```

In my previous approach, I did not declare `fastgltf_module` CMake target in the project itself and, but in the user project(`fastgltf-module-test`). Since two targets are separated, I can enable the `std` module between two targets, and successfully build. I think this approach is the most used among the module support projects like [Vulkan-Hpp](https://github.com/stripe2933/fastgltf-module-test/actions/runs/9058919976/job/24885513529), [glm](https://github.com/g-truc/glm/blob/master/glm/glm.cppm) and other: declaring module target is the responsibility of the user.

Another reason I worried about is: will package managers like vcpkg maintain module target? For same case, the library `fmt` cannot be used with module in vcpkg, because vcpkg disabled the related CMake options.

Of course there is no general rule for it. As module user population grow up, many CMake based project will expose the module target, so the choice is up to you.

Anyway, could you share me your local module build project? I'm curious how `std` target only affects to `fastgltf_module` but not `fastgltf` target.

spnda:
> Anyway, could you share me your local module build project? I'm curious how std target only affects to fastgltf_module but not fastgltf target.

I just used your `fastgltf-module-test` repository locally, but didn't use the `import std` method with your CMake library. I use Clang 18.1 locally with CMake 3.28.1. I don't use `import std` locally since your library gives this error when configuring:
```
CMake Error at /Users/sean/Applications/CLion.app/Contents/bin/cmake/mac/aarch64/share/cmake-3.28/Modules/ExternalProject.cmake:1672 (message):
  Do not know how to extract '/modules/c++/v1/' -- known types are: .7z,
  .tar, .tar.bz2, .tar.gz, .tar.xz, .tbz2, .tgz, .txz and .zip
```

* If the target is declared before *your* `std` library it won't work, yes. Though that looks more like a limitation of how your library works, and can be easily fixed by just importing fastgltf before including your library. Those can both be documented in the README of your library. I don't think CMake's own functionality for importing the sodlib module will have this limitation, but correct me if I'm wrong.

* The errors about `install` you're getting are just because your `std` library does not provide any install commands. If someone wants to install fastgltf, CMake needs to know how to install your `std` library, too. Not sure how to go about that, though, since I think it's statically linked into any executable/library anyway?

spnda:
Just tested locally with a CMake 3.30 nightly release, and using `import std` works fine on MSVC. The interface will hopefully change when CMake 3.30 releases, but it should show that the module target order is not relevant to CMake.
```cmake
cmake_minimum_required(VERSION 3.29.20240513)

# Needs to happen before the project declaration, or C++ language support.
set(CMAKE_EXPERIMENTAL_CXX_IMPORT_STD "0e5b6991-d74f-4b3d-a41c-cf096e0b2508")

project(fastgltf-module-test LANGUAGES CXX)

set(CMAKE_CXX_STANDARD 23)
set(CMAKE_CXX_MODULE_STD 1) # Required for 'import std'

option(FASTGLTF_USE_STD_MODULE "" ON)
add_subdirectory("E:/git/spnda/fastgltf/" fastgltf)

add_executable(fastgltf-module-test main.cpp)
target_link_libraries(fastgltf-module-test PRIVATE fastgltf::module)
```

stripe2933:
> Just tested locally with a CMake 3.30 nightly release, and using `import std` works fine on MSVC. The interface will hopefully change when CMake 3.30 releases, but it should show that the module target order is not relevant to CMake.

My library is just a stopgap implementation for CMake 3.30's feature, therefore it would be my library's problem. I'll investigate it.

All changes look good to me.

stripe2933:
> Just tested locally with a CMake 3.30 nightly release, and using `import std` works fine on MSVC. The interface will hopefully change when CMake 3.30 releases, but it should show that the module target order is not relevant to CMake.

Ok, my library is just a stopgap implementation for CMake 3.30's feature, so it may result incorrect behavior. I'll investigate it.

---

It looks like Clang does not accept `export` for redeclared classes. [CI error](https://github.com/stripe2933/fastgltf-module-test/actions/runs/9092229281/job/24988492293) MSVC works well.

spnda (approved):

spnda:
@stripe2933 Thank you very much for proposing this and the initial changes, and the last fixes. I've merged this now.

PR #64, opened 2024-06-21
Add: Animation support to Exporter; Fix: Vector source not allowed as embedded buffer

Deweh:
This PR adds support for animations to the `Exporter`, and also fixes an issue with the `Exporter` where a `sources::Vector` buffer would not pass the check for using an embedded buffer, even though documentation states it should be usable for an embedded buffer.

### DIFF
--- a/src/fastgltf.cpp
+++ b/src/fastgltf.cpp
@@ -4192,6 +4192,79 @@ void fg::Exporter::writeAccessors(const Asset& asset, std::string& json) {
 	json += ']';
 }
 
+void fg::Exporter::writeAnimations(const Asset& asset, std::string& json)
+{
+	if (asset.animations.empty())
+		return;
+	if (json.back() == ']' || json.back() == '}')
+		json += ',';
+
+	json += R"("animations":[)";
+	for (auto it = asset.animations.begin(); it != asset.animations.end(); ++it) {
+		json += '{';
+
+		json += R"("channels":[)";
+		for (auto ci = it->channels.begin(); ci != it->channels.end(); ++ci) {
+			json += "{";
+			json += R"("sampler":)" + std::to_string(ci->samplerIndex) + ",";
+			json += R"("target":{)";
+			if (ci->nodeIndex.has_value()) {
+				json += R"("node":)" + std::to_string(ci->nodeIndex.value()) + ",";
+			}
+			json += R"("path":")";
+			switch (ci->path) {
+			case fg::AnimationPath::Translation:
+				json += "translation";
+				break;
+			case fg::AnimationPath::Rotation:
+				json += "rotation";
+				break;
+			case fg::AnimationPath::Scale:
+				json += "scale";
+				break;
+			case fg::AnimationPath::Weights:
+				json += "weights";
+				break;
+			}
+			json += "\"}}";
+
+			if (uabs(std::distance(it->channels.begin(), ci)) + 1 < it->channels.size())
+				json += ',';
+		}
+		json += "],";
+
+		json += R"("name":")" + it->name + ",";
+
+		json += R"("samplers":[)";
+		for (auto si = it->samplers.begin(); si != it->samplers.end(); ++si) {
+			json += "{";
+			json += R"("input":)" + std::to_string(si->inputAccessor) + ",";
+			json += R"("interpolation":")";
+			switch (si->interpolation) {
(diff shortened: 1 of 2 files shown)


spnda:
Did you intend to target this for the 0.7.x branch? I intend to release 0.8 rather soon and only released 0.7.2 recently for bug fixes, since some people don't want to deal with the breaking changes yet. If not I can switch this to go into the main branch instead.

Otherwise, this looks good (I thought the Exporter supported everything already lol) and good catch about the embedded buffer. I reworked `Vector` into `Array`, but then re-added `Vector` since it's a more suitable option for some use cases, and must've forgot re-adding it here. Perhaps there are other places where I missed it...

src/fastgltf.cpp
```diff
@@ -4192,6 +4192,79 @@ void fg::Exporter::writeAccessors(const Asset& asset, std::string& json) {
...
+				break;
+			}
+			json += "\"}}";
+
+			if (uabs(std::distance(it->channels.begin(), ci)) + 1 < it->channels.size())
+				json += ',';
+		}
+		json += "],";
+
+		json += R"("name":")" + it->name + "\",";
```
spnda, line 4236: This should probably first check `if (!it->name.empty())`. Also, the `name` field should probably be written last, as I do this with all other objects, too, and want to keep the order as consistent as possible. The spec also always puts the name, extensions, and extras fields as the last properties, in that order.

src/fastgltf.cpp
```diff
@@ -4192,6 +4192,79 @@ void fg::Exporter::writeAccessors(const Asset& asset, std::string& json) {
...
+			case fastgltf::AnimationInterpolation::Linear:
+				json += "LINEAR";
+				break;
+			case fastgltf::AnimationInterpolation::Step:
+				json += "STEP";
+				break;
+			case fastgltf::AnimationInterpolation::CubicSpline:
+				json += "CUBICSPLINE";
+				break;
+			}
```
spnda, line 4253: Since the default for the interpolation property is `Linear`, we should skip serializing that if `si->interpolation` is `Linear`. I do this everywhere else, too, to keep JSON sizes low and increase parsing speeds.

src/fastgltf.cpp
```diff
@@ -4192,6 +4192,79 @@ void fg::Exporter::writeAccessors(const Asset& asset, std::string& json) {
...
+			json += "\",";
+			json += R"("output":)" + std::to_string(si->outputAccessor);
+			json += "}";
```
spnda, line 4256: These two strings could be concatenated into one, to reduce the amount of string resize operations. Also, the `}` is a single character and could be converted to one.

spnda (changes requested):

Deweh:
> Did you intend to target this for the 0.7.x branch? I intend to release 0.8 rather soon and only released 0.7.2 recently for bug fixes, since some people don't want to deal with the breaking changes yet. If not I can switch this to go into the main branch instead.

I don't PR very often, so I'm not familiar with the proper procedure for going about these things, apologies. I just pushed to the 0.7.x branch since that's the branch a project of mine was on when I realized I needed these changes.

I'll push those requested changes soon.

spnda:
> I don't PR very often, so I'm not familiar with the proper procedure for going about these things, apologies.
No worries, this is more just about what I have intended for the future, you can't really know that.

Since this is technically a bugfix, we could still PR this onto the 0.7.x branch, and then I'll bring the commit over to the main branch, since there might be some conflicts and then I can just resolve those. I'll wait on the additional changes, test them myself, and then I'll merge.

Deweh:
> > I don't PR very often, so I'm not familiar with the proper procedure for going about these things, apologies.
> > No worries, this is more just about what I have intended for the future, you can't really know that.
> 
> Since this is technically a bugfix, we could still PR this onto the 0.7.x branch, and then I'll bring the commit over to the main branch, since there might be some conflicts and then I can just resolve those. I'll wait on the additional changes, test them myself, and then I'll merge.

Sounds good. I've pushed the changes, and also added the extrasWriteCallback before the name to match the functions for the other types.

spnda (approved):

spnda:
Thank you, I've merged this onto the `v0.7.x` branch and cherry-picked it onto the `main` branch as 65e1fec04eccff1939f68f09aaca35152104ea30.

PR #73, opened 2024-09-11
Update gl_viewer.cpp

cmd05:
Use binding parameter for uniform block instead of location parameter. 

https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf 
Section 4.4 - Layout Qualifiers mentions `location =` as not applicable for blocks

Error:
```
GL Renderer: AMD Radeon(TM) Graphics
GL Version: 4.6.0 Compatibility Profile Context 22.20.44.37.230215
Shader compilation error: 
ERROR: 0:13: 'location' : cannot apply to uniform or buffer block
ERROR: 0:13: 'location' : overlapping use of location 0
ERROR: 2 compilation errors.  No code generated.
```


spnda (approved):
I don't much about OGL to be honest, but @deccer confirmed that this change is correct.

PR #79, opened 2024-11-16
Fix namespacing for DirectXMath element traits

MattGuerrette:
This commit corrects the namespacing for the DirectXMath element traits as some of the XM types come from the DirectX::PackedVector namespace.


MattGuerrette:
FYI, I made these changes while evaluating this library. I have not personally tested the use of these, but this change will at least fix compilation.

spnda:
Uh one sec please, I tried to add another commit with another header include for the packed vectors and I somehow broke something with Git.

spnda:
Test

spnda:
Uh, could you please push that commit back onto that branch, so that I can re-open this PR?

PR #80, opened 2024-11-16
Fix namespacing for DirectXMath element traits

MattGuerrette:
This commit corrects the namespacing for the DirectXMath element traits as some of the XM types come from the DirectX::PackedVector namespace.


spnda:
Great, thank you. Sorry about that...
I added a second commit that also includes `DirectXPackedVector.h` if its available, since the `PackedVector` namespace is only defined there. Also added the `XMUBYTEN` and corrected the `XMSHORTN` specializations.

PR #84, opened 2024-12-12
Added new error message 

wpsimon09:
```c++
void fg::GltfDataBuffer::read(void *ptr, std::size_t count) {
```

Following function did not have error handling and while I was attempting to use the library SEGV occurred on this function. I have added safety `if(buffer != nullptr)` check to prevent this.

Additionally I have also added new entry to the `enum class Error` that signalises that reading from the buffer failed. Lastly  `getErrorName` and `getErrorMessage` were updated so that they reflect new entry in the `enum` .


spnda:
This doesn't make sense since the check is just for `buffer` being `nullptr`, which shouldn't happen anyway. This means that you were passing the `GltfDataBuffer` constructor an invalid buffer, which should probably trigger an assert. The error message for the given circumstances also does not make sense at all.

PR #85, opened 2024-12-15
Include and use GNUInstallDirs in cmake

pclouds:
This installs libfastgltf.a in the expected directory. On many Linux distros this will be /usr/lib64, not /usr/lib. And this is important when the distro supports both 32 and 64 bits (32 libs go to /usr/lib)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,7 @@ else()
     set(FASTGLTF_COMPILE_TARGET cxx_std_17)
 endif()
 
+include(GNUInstallDirs)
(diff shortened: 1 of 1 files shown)


spnda:
First of all sorry for the late response. Do you know if this has any impact on things like vcpkg or conan? It just feels a little weird it being called `GNUInstallDirs`. Otherwise I'd be fine with this.

spnda:
Ok sure, I think I'll just merge this and see if anybody complains. I think most of the time the folders would resolve to the same as before anyway.

PR #91, opened 2025-03-11
Fix CMake installation issues

marovira:
## What is Being Fixed
This pull request fixes two issues I noticed when I installed the library for development:
1. The package version manifest isn't being installed.
2. There is no way to install debug and release builds of the library without manually forcing CMake to do so.

### Issue 1
A common practice when using `find_package` in CMake is to pin the minimum required version so there are no conflicts between the user code and libraries. This is generally achieved by doing `find_package(<name> <version>)`. This relies on the existence of a `<name>configVersion.cmake` file being installed into `lib/cmake/<name>`. With the current code, this file is never generated and so never installed, preventing users from requesting a specific version of `fastgltf`.

### Issue 2
When building in Windows with Visual Studio, it is common practice to build and install two sets of binaries: one for debug and one for release builds. In user code, CMake will then automatically link the correct binary for the appropriate build. In the current CMake code, there is no way to differentiate between the debug and release builds. This results in the following: if debug is built and installed first, followed by release, the debug library will be overwritten as both debug and release are called `fastgltf.lib`. Using this library in debug builds triggers compiler errors due to mismatched optimisation levels. 

This can be worked around by manually setting `CMAKE_DEBUG_POSTFIX=d` when building the library, which is error prone and is better set by the library itself. 

I have verified that both issues are solved with this pull request, and based on past experience with other libraries, doesn't negatively impact other platforms. I think this won't be an issue with `vcpkg` but I've never used it before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,7 @@ else()
 endif()
 
 include(GNUInstallDirs)
+include(CMakePackageConfigHelpers)
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/add_source_directory.cmake)
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/compilers.cmake)
 
@@ -140,9 +141,20 @@ elseif(FASTGLTF_ENABLE_CPP_MODULES)
     message(WARNING "FASTGLTF_ENABLE_CPP_MODULES is ON but compiler does not support them")
 endif()
 
+write_basic_package_version_file(
+    "${PROJECT_BINARY_DIR}/fastgltfConfigVersion.cmake"
+    VERSION ${PACKAGE_VERSION}
+    COMPATIBILITY SameMajorVersion
+)
+
 install(
     FILES ${FASTGLTF_HEADERS}
     DESTINATION include/fastgltf
+)
+
+install(FILES
+    "${CMAKE_CURRENT_BINARY_DIR}/fastgltfConfigVersion.cmake"
+    DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/fastgltf
 )
 
 install(


CMakeLists.txt
```diff
@@ -110,6 +111,7 @@ target_compile_definitions(fastgltf PUBLIC "FASTGLTF_USE_CUSTOM_SMALLVECTOR=$<BO
 target_compile_definitions(fastgltf PUBLIC "FASTGLTF_ENABLE_DEPRECATED_EXT=$<BOOL:${FASTGLTF_ENABLE_DEPRECATED_EXT}>")
 target_compile_definitions(fastgltf PUBLIC "FASTGLTF_DISABLE_CUSTOM_MEMORY_POOL=$<BOOL:${FASTGLTF_DISABLE_CUSTOM_MEMORY_POOL}>")
 target_compile_definitions(fastgltf PUBLIC "FASTGLTF_USE_64BIT_FLOAT=$<BOOL:${FASTGLTF_USE_64BIT_FLOAT}>")
+set_target_properties(fastgltf PROPERTIES DEBUG_POSTFIX "d")
```
spnda, line 114: After looking at https://cmake.org/cmake/help/latest/guide/tutorial/Packaging%20Debug%20and%20Release.html it seems like this is something the user would do. I would, however, agree that the following would be perfectly acceptable for a library:
```cmake
set_target_properties(fastgltf PROPERTIES DEBUG_POSTFIX ${CMAKE_DEBUG_POSTFIX})
```

I feel like hardcoding this could cause more problems than it solves, though I don't know any specifics here to be perfectly honest.

marovira: This change is effectively the same as the current setup: the user has to define `CMAKE_DEBUG_POSTFIX` for debug and release builds to co-exist as CMake does not provide a default value for `CMAKE_DEBUG_POSTFIX`. I do know that it is relatively "common" (as common as things can be in C++/CMake without a real standard) for libraries to define this themselves (some use `d`, others use `_debug`, but the overall idea is the same) to facilitate this exact workflow. But there's also several libraries that don't do this, so I can revert this change if you're unsure. Just figured I'd throw it in there to make life easier :)

spnda: After reading https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIG_POSTFIX.html this would not be necessary anyway, since the `DEBUG_POSTFIX` is automatically initialized to `CMAKE_DEBUG_POSTFIX`, if it is defined, anyway.

I tried searching for other examples of this and found these some libraries hardcoding `CMAKE_CONFIG_POSTFIX` to `"-d"`, whereas other people are strictly against the library including any such code besides inheriting the property.

To be honest I feel like the second commit can be removed for now. I like the first commit, but I'm not sure about the second and I feel like it should be a separate discussion and I think ultimately it won't be included since users who want this behavior should probably just be setting it explicitly when installing the library.

marovira: Second commit has been removed.

spnda (approved):

PR #94, opened 2025-04-23
Fix fastgltf::URIView::fspath() error when path contains non-ascii character in Windows.

stripe2933:
Reinterpreted the returned value of `path()` as `std::u8string_view` to use the platform-specific encoded `std::filesystem::path`.

I think the better option would be change the return type of `fastgltf::URIView::path()` as `char8_t`, but it is C++20 feature and therefore will break the backward compatibility.

### DIFF
--- a/src/fastgltf.cpp
+++ b/src/fastgltf.cpp
@@ -518,7 +518,16 @@ std::string_view fg::URIView::fragment() const noexcept { return _fragment; }
 fs::path fg::URIView::fspath() const {
 	if (!isLocalPath())
 		return {};
+#ifdef _WIN32
+#if __cplusplus >= 202002L
+	const std::string_view charPath = path();
+	return { std::u8string_view { reinterpret_cast<char8_t>(charPath.data()), charPath.size() } };
+#else
+	return fs::u8path(path());
+#endif
+#else
 	return { path() };
+#endif
 }
 
 bool fg::URIView::valid() const noexcept {


spnda:
I guess the Unicode tests never checked the `fspath` function.
https://github.com/spnda/fastgltf/blob/eebda658317cd907d6cdf9c9ed7e40b96e8453dd/tests/basic_test.cpp#L495-L517

I think it would make sense to have that test also cover more functionality of URI to be sure everything works together with Unicode, seeing as the tests never caught this problem.

The Unicode support in C++ is generally quite weird, because even in post-C++20 code people continue to use `char`/`std::string` together with UTF-8/Unicode, mostly because C APIs will not be compatible otherwise. I'd rather stick to  just using the original types in all user-facing code and just quietly assume everything is UTF-8.

src/fastgltf.cpp
```diff
@@ -518,7 +518,16 @@ std::string_view fg::URIView::fragment() const noexcept { return _fragment; }
 fs::path fg::URIView::fspath() const {
 	if (!isLocalPath())
 		return {};
+#ifdef _WIN32
+#if __cplusplus >= 202002L
+	const std::string_view charPath = path();
+	return { std::u8string_view { reinterpret_cast<char8_t>(charPath.data()), charPath.size() } };
```
spnda, line 524: isn't this casting a `char*` to a non-pointer `char8_t`?

src/fastgltf.cpp
```diff
@@ -518,7 +518,16 @@ std::string_view fg::URIView::fragment() const noexcept { return _fragment; }
 fs::path fg::URIView::fspath() const {
 	if (!isLocalPath())
 		return {};
+#ifdef _WIN32
+#if __cplusplus >= 202002L
```
spnda, line 522: 1. `FASTGLTF_CPP_20` is the more appropriate, platform-independent macro that should be used here.
2. Maybe we should check for [`__cpp_char8_t`](https://en.cppreference.com/w/cpp/feature_test#cpp_char8_t) instead?

src/fastgltf.cpp
```diff
@@ -518,7 +518,16 @@ std::string_view fg::URIView::fragment() const noexcept { return _fragment; }
...
+	const std::string_view charPath = path();
+	return { std::u8string_view { reinterpret_cast<char8_t>(charPath.data()), charPath.size() } };
+#else
+	return fs::u8path(path());
+#endif
+#else
 	return { path() };
```
spnda, line 529: Are you sure that the current implementation works as expected on Linux/MacOS? The URI code uses `u8string_view` elsewhere to fix this issue, and I have a feeling that it could be the same here. Have you tested this?

spnda (changes requested):

src/fastgltf.cpp
stripe2933: Both macOS and Linux are POSIX system, therefore I believe it is okay to use `char` based constructor.

However, for code simplicity it would be better to use `char8_t` constructor for both POSIX and non-POSIX system (former will do nothing for path conversion). Fixed.

spnda (approved):

PR #95, opened 2025-04-29
Khr physics rigid bodies

DethRaid:
Adds support for the draft [KHR_physics_rigid_bodies](https://github.com/KhronosGroup/glTF/pull/2424) extension, and [KHR_implicit_shapes](https://github.com/KhronosGroup/glTF/pull/2370)

The import code is decently well-tested and used in my game, the export code is a bit rougher

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -99,10 +99,17 @@ When this ``BOOL`` option is set to ``YES`` fastgltf will use `corrosion`_, whic
 to link against the `gltf-rs`_ Rust library for comparison within the benchmarks.
 Note that this option has no effect when ``FASTGLTF_ENABLE_TESTS`` is set to ``NO``.
 
+``FASTGLTF_ENABLE_IMPLICIT_SHAPES``
+--------------------------------------------
+
+This option enables support for the draft KHR_implicit_shapes <https://github.com/KhronosGroup/glTF/pull/2370> extension
+
 ``FASTGLTF_ENABLE_KHR_PHYSICS_RIGID_BODIES``
 --------------------------------------------
 
 This option enables support for the draft KHR_physics_rigid_bodies <https://github.com/KhronosGroup/glTF/pull/2424> extension
+
+KHR_physics_rigid_bodies relies on KHR_implicit_shapes, so this option implicitly enables `FASTGLTF_ENABLE_IMPLICIT_SHAPES`
 
 Parsing options
 ===============
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -99,6 +99,11 @@ When this ``BOOL`` option is set to ``YES`` fastgltf will use `corrosion`_, whic
 to link against the `gltf-rs`_ Rust library for comparison within the benchmarks.
 Note that this option has no effect when ``FASTGLTF_ENABLE_TESTS`` is set to ``NO``.
 
+``FASTGLTF_ENABLE_KHR_PHYSICS_RIGID_BODIES``
+--------------------------------------------
+
+This option enables support for the draft KHR_physics_rigid_bodies <https://github.com/KhronosGroup/glTF/pull/2424> extension
+
 Parsing options
 ===============
 
--- a/include/fastgltf/types.hpp
+++ b/include/fastgltf/types.hpp
@@ -301,6 +301,27 @@ namespace fastgltf {
     FASTGLTF_ASSIGNMENT_OP_TEMPLATE_MACRO(Category, Category, &)
     FASTGLTF_UNARY_OP_TEMPLATE_MACRO(Category, ~)
     // clang-format on
+
+
+#if FASTGLTF_ENABLE_KHR_PHYSICS_RIGID_BODIES
+	FASTGLTF_EXPORT enum class CombineMode : std::uint8_t {
+        Average,
+		Minimum,
+		Maximum,
+		Multiply,
+		Invalid,
+    };
+
+	FASTGLTF_EXPORT enum class DriveType : std::uint8_t {
+	    Linear,
+		Angular
+	};
+
+	FASTGLTF_EXPORT enum class DriveMode : std::uint8_t {
+	    Force,
+		Acceleration
+	};
+#endif
 #pragma endregion
 
 #pragma region ConversionFunctions
@@ -479,6 +500,41 @@ namespace fastgltf {
 				return "";
 		}
 	}
+
+#if FASTGLTF_ENABLE_KHR_PHYSICS_RIGID_BODIES
+	constexpr auto getFrictionCombineType(std::string_view name) noexcept {
+		assert(!name.empty());
+	    if(name[0] == 'a') {
+			return CombineMode::Average;
+	    }
+
+		switch(name[1]) {
+	    case 'i':
+			return CombineMode::Minimum;
+
+	    case 'a':
+			return CombineMode::Maximum;
+
+	    case 'u':
+			return CombineMode::Multiply;
+		}
+
+		return CombineMode::Invalid;
+	}
+
+	static constexpr std::array<std::string_view, 4> frictionCombineNames{
+        "average",
+        "minimum",
+        "maximum",
+        "multiply"
+	};
+
+	constexpr std::string_view getFrictionCombineName(CombineMode frictionCombine) noexcept {
+		static_assert(std::is_same_v<std::underlying_type_t<CombineMode>, std::uint8_t>);
+		const auto idx = to_underlying(frictionCombine) & 0x3;
+		return frictionCombineNames[idx];
+	}
+#endif
 #pragma endregion
 
 #pragma region Containers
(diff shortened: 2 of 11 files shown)


src/fastgltf.cpp
```diff
@@ -3842,10 +3857,334 @@ fg::Error fg::Parser::parseTextures(simdjson::dom::array& textures, Asset& asset
...
+		if (collisionFilterValue.get_object().get(collisionFilterObject) != SUCCESS) FASTGLTF_UNLIKELY {
+			return Error::InvalidGltf;
+		}
+
+		dom::array collisionSystems;
+		if (auto error = collisionFilterObject["collisionSystems"].get_array().get(collisionSystems); error == SUCCESS) FASTGLTF_LIKELY {
+			collisionFilter.collisionSystems.reserve(collisionSystems.size());
+			for (auto systemNameValue : collisionSystems) {
+				std::string_view systemName;
+				systemNameValue.get_string().get(systemName);
```
spnda, line 3938: no error checks

src/fastgltf.cpp
```diff
@@ -3842,10 +3857,334 @@ fg::Error fg::Parser::parseTextures(simdjson::dom::array& textures, Asset& asset
...
+		if(collisionFilterObject["collideWithSystems"].error() == SUCCESS && collisionFilterObject["notCollideWithSystems"].error() == SUCCESS) {
+			return Error::InvalidGltf;
+		}
+
+		dom::array collideWithSystems;
+		if (auto error = collisionFilterObject["collideWithSystems"].get_array().get(collideWithSystems); error == SUCCESS) FASTGLTF_LIKELY {
+			collisionFilter.collideWithSystems.reserve(collideWithSystems.size());
+			for(auto systemNameValue : collideWithSystems) {
+				std::string_view systemName;
+				systemNameValue.get_string().get(systemName);
```
spnda, line 3955: no error checks

src/fastgltf.cpp
```diff
@@ -3842,10 +3857,334 @@ fg::Error fg::Parser::parseTextures(simdjson::dom::array& textures, Asset& asset
...
+		} else if (error != NO_SUCH_FIELD) {
+			return Error::InvalidGltf;
+		}
+
+		dom::array notCollideWithSystems;
+		if (auto error = collisionFilterObject["notCollideWithSystems"].get_array().get(notCollideWithSystems); error == SUCCESS) FASTGLTF_LIKELY {
+			collisionFilter.notCollideWithSystems.reserve(notCollideWithSystems.size());
+			for (auto systemNameValue : notCollideWithSystems) {
+				std::string_view systemName;
+				systemNameValue.get_string().get(systemName);
```
spnda, line 3967: no error checks

spnda (commented):
First of all big thanks. Don't have much to say yet, but one thing I will mention is that the `if`/`else`/`for` style is wildly inconsistent, most often with no space after an `if`/`for`, and sometimes with a newline before an `else`. While I generally don't care about the style contributors use (you'll find a few different styles throughout the codebase), I'd still like it to be as consistent as possible, and at best stick to the general codestyle.

src/fastgltf.cpp
```diff
@@ -5711,13 +6258,368 @@ void fg::Exporter::writeTextures(const Asset& asset, std::string& json) {
...
+				if (limit.min) {
+					json += R"("min":)" + std::to_string(*limit.min) + ",";
+				}
+				if (limit.max) {
+					json += R"("max":)" + std::to_string(*limit.max) + ",";
+				}
+				if (limit.stiffness) {
+					json += R"("stiffness":)" + std::to_string(*limit.stiffness) + ",";
+				}
+			    json += R"("damping":)" + std::to_string(limit.damping);
```
spnda, line 6524: These should all be `to_string_fp` since `std::to_string(float)` doesn't have round-trip guarantees.

src/fastgltf.cpp
```diff
@@ -5711,13 +6258,368 @@ void fg::Exporter::writeTextures(const Asset& asset, std::string& json) {
...
+					json += R"("acceleration")";
+					break;
+				}
+
+				json += R"(,"axis":)" + std::to_string(drive.axis)
+			        + R"(,"maxForce":)" + std::to_string(drive.maxForce)
+			        + R"(,"positionTarget":)" + std::to_string(drive.positionTarget)
+			        + R"(,"velocityTarget":)" + std::to_string(drive.velocityTarget)
+			        + R"(,"stiffness":)" + std::to_string(drive.stiffness)
+			        + R"(,"damping":)" + std::to_string(drive.damping);
```
spnda, line 6568: Also `to_string_fp` here.

src/fastgltf.cpp
```diff
@@ -5711,13 +6258,368 @@ void fg::Exporter::writeTextures(const Asset& asset, std::string& json) {
...
+		    break;
+		case ShapeType::Capsule:
+		    {
+			    const auto& capsule = std::get<CapsuleShape>(shape.shape);
+			    json += R"("type":"capsule","capsule":{"height":)" + std::to_string(capsule.height) + R"(,"radiusBottom":)" + std::to_string(capsule.radiusBottom) + R"(,"radiusTop":)" + std::to_string(capsule.radiusTop) + "}";
+		    }
+            break;
+        case ShapeType::Invalid:
+            break;
+        }
```
spnda, line 6302: Also `to_string_fp`

src/fastgltf.cpp
```diff
@@ -5502,6 +5960,95 @@ void fg::Exporter::writeNodes(const Asset& asset, std::string& json) {
...
+				        + R"(],"gravityFactor":[)" + to_string_fp(motion.gravityFactor) + "}";
+				}
+
+				if (it->physicsRigidBody->trigger.has_value()) {
+					if (json.back() != '{') json += ',';
+					const auto& trigger = *it->physicsRigidBody->trigger;
+					json += R"("trigger":{)";
+					std::visit(visitor{
+	            [&](auto&) {
+							// Is this needed?
```
spnda, line 5992: Nope, and this should probably use `fastgltf::visit_exhaustive`, which checks if you have all the overloads specified correctly, and aren't missing any, have wrong const-ness, or ... by accident.

src/fastgltf.cpp
```diff
@@ -5502,6 +5960,95 @@ void fg::Exporter::writeNodes(const Asset& asset, std::string& json) {
...
+
+#if FASTGLTF_ENABLE_KHR_PHYSICS_RIGID_BODIES
+			if (it->physicsRigidBody.has_value()) {
+				if (json.back() != '{') json += ',';
+				json += R"("KHR_physics_rigid_body":{)";
+				if (it->physicsRigidBody->motion.has_value()) {
+					const auto& motion = *it->physicsRigidBody->motion;
+					json += R"("motion":{"isKinematic":)" + std::to_string(motion.isKinematic);
+					if (motion.mass.has_value()) {
+						json += R"(,"mass":)" + std::to_string(*motion.mass);
```
spnda, line 5972: Also `to_string_fp`

spnda (changes requested):
Went over this quickly, most of it looks good. I'll do an in depth review when you have added some testing code, so I can test locally too. Perhaps I should really add some kind of fuzzer, since I'd assume there's only a handful of available assets with this extension.

Well apart from the comments I'd like to mention that your code doesn't use `FASTGLTF_UNLIKELY` and `FASTGLTF_LIKELY` much (except for the parts I guess you copy & pasted). Might look annoying, but I found them to make parsing a few percents quicker due to better codegen, in the cases where no errors are returned.

Also for the exporting code (sorry you had to go through that mess), single characters should always be `'x'` and not a string.

src/fastgltf.cpp
```diff
@@ -3857,12 +3860,164 @@ fg::Error fg::Parser::parseTextures(simdjson::dom::array& textures, Asset& asset
...
+				return Error::InvalidGltf;
+		    }
+
+			dom::array sizeArray;
+			if (error = boxObject["size"].get_array().get(sizeArray); error == SUCCESS) {
+			    if (sizeArray.size() != 3) {
+					return Error::InvalidGltf;
+			    }
+
+				auto box = BoxShape{};
```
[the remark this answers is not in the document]
DethRaid: Oh, I didn't realize `optional` had an `emplace` method

src/fastgltf.cpp
```diff
@@ -3605,6 +3659,18 @@ fg::Error fg::Parser::parseNodes(simdjson::dom::array& nodes, Asset& asset) {
...
+
+#if FASTGLTF_ENABLE_KHR_PHYSICS_RIGID_BODIES
+			if (hasBit(config.extensions, Extensions::KHR_physics_rigid_bodies)) {
+				dom::object physicsRigidBodiesObject;
+				if (extensionsObject[extensions::KHR_physics_rigid_bodies].get_object().get(physicsRigidBodiesObject) == SUCCESS) FASTGLTF_LIKELY{
+					const auto rigidBodyError = parsePhysicsRigidBody(physicsRigidBodiesObject, node);
+					if (rigidBodyError != Error::None) {
+						return rigidBodyError;
+					}
+			    }
```
spnda, line 3671: Missing a check if simdjson returned `NO_SUCH_FIELD` here.

src/fastgltf.cpp
```diff
@@ -3842,180 +3908,938 @@ fg::Error fg::Parser::parseTextures(simdjson::dom::array& textures, Asset& asset
...
+						box.size[curIndex] = static_cast<num>(value);
+					} else {
+						return Error::InvalidGltf;
+					}
 
-fg::Expected<fg::Asset> fg::Parser::loadGltf(GltfDataGetter& data, fs::path _directory, Options _options, Category categories) {
-    auto type = fastgltf::determineGltfFileType(data);
+					curIndex++;
+				}
+			}
```
spnda, line 3971: Missing error check & wrong usage of likely

src/fastgltf.cpp
```diff
@@ -3842,180 +3908,938 @@ fg::Error fg::Parser::parseTextures(simdjson::dom::array& textures, Asset& asset
...
-    // If we never have to load the files ourselves, we're fine with the directory being invalid/blank.
-    if (std::error_code ec; hasBit(options, Options::LoadExternalBuffers) && (!fs::is_directory(directory, ec) || ec)) {
-	    return Error::InvalidPath;
-    }
+			double radiusTop;
+			if (error = cylinderObject["radiusTop"].get_double().get(radiusTop); error == SUCCESS) FASTGLTF_LIKELY {
+				cylinder.radiusTop = static_cast<num>(radiusTop);
+			} else if (error != NO_SUCH_FIELD) {
+				return Error::InvalidGltf;
+			}
```
spnda, line 4036: Wrong usage of likely

docs/options.rst
```diff
@@ -99,6 +99,11 @@ When this ``BOOL`` option is set to ``YES`` fastgltf will use `corrosion`_, whic
 to link against the `gltf-rs`_ Rust library for comparison within the benchmarks.
 Note that this option has no effect when ``FASTGLTF_ENABLE_TESTS`` is set to ``NO``.
 
+``FASTGLTF_ENABLE_KHR_PHYSICS_RIGID_BODIES``
+--------------------------------------------
+
+This option enables support for the draft KHR_physics_rigid_bodies <https://github.com/KhronosGroup/glTF/pull/2424> extension
+
```
spnda, line 106: Shouldn't this also list the option for the `KHR_implicit_shapes` extension?

spnda (changes requested):
- I know the usage of likely/unlikely is sometimes wrong in my code, too, but it would be best to get it right here.
- There's a few issues around indentation, specifically inconsistency between tabs and spaces. I want to slowly switch to using tabs, without creating a commit changing thousands of lines.
- There's probably still a few things I missed, but the fact the tests succeed everywhere is a good sign. Will need to review the tests in a bit more detail soon, so I know that everything is covered.

tests/extension_tests.cpp
```diff
@@ -480,3 +480,276 @@ TEST_CASE("Extension KHR_materials_variant", "[gltf-loader]") {
...
+			REQUIRE(false);
+		},
+		[](const fastgltf::CapsuleShape& capsule) {
+			REQUIRE(false);
+		},
+		[](const fastgltf::CylinderShape& cylinder) {
+			REQUIRE(false);
+		}
+		},
+		box0.shape);
```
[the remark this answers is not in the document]
DethRaid: Hmm, I suppose we don't need `type`. I initially put it int he struct because it was in the glTF spec, but you're right that it's redundant information

I did an exhaustive visit because I wanted the test to error if the deserialized shape was anything other than a BoxShape - but perhaps there's a better way to express that

tests/extension_tests.cpp
```diff
@@ -480,3 +480,276 @@ TEST_CASE("Extension KHR_materials_variant", "[gltf-loader]") {
...
+			REQUIRE(node.physicsRigidBody->collider->geometry.shape.has_value());
+			REQUIRE(node.physicsRigidBody->collider->geometry.shape == 0);
+			REQUIRE(!node.physicsRigidBody->collider->geometry.node.has_value());
+			REQUIRE(node.physicsRigidBody->collider->physicsMaterial == 0);
+			REQUIRE(node.physicsRigidBody->collider->collisionFilter == 0);
+		}
+
+		REQUIRE(!node.physicsRigidBody->trigger.has_value());
+		REQUIRE(!node.physicsRigidBody->joint.has_value());
+	}
```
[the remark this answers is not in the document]
DethRaid: The branches made my static analyzer stop complaining about "unchecked optional access", but I can remove them

include/fastgltf/types.hpp
```diff
@@ -1888,6 +1988,266 @@ namespace fastgltf {
...
+
+        /**
+         * An array of strings representing the systems which this node can _not_ collide with
+         */
+        FASTGLTF_FG_PMR_NS::MaybeSmallVector<FASTGLTF_STD_PMR_NS::string> notCollideWithSystems;
+
+        /**
+         * An array of strings representing the systems which this node can collide with
+         */
+        FASTGLTF_FG_PMR_NS::MaybeSmallVector<FASTGLTF_STD_PMR_NS::string> collideWithSystems;
```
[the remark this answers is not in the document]
DethRaid: Yes, these are arbitrary names for systems. I agree it's a little weird. I'd call them 'collision layers" personally, since that's what I've seen in Unreal and Unity, and numbers would be perfectly fine... but all well. This is how KHR_physics_rigid_bodies defines collision filters, see https://github.com/eoineoineoin/glTF/tree/rigidBodyMerge/extensions/2.0/Khronos/KHR_physics_rigid_bodies#collision-filtering and https://github.com/eoineoineoin/glTF/blob/rigidBodyMerge/extensions/2.0/Khronos/KHR_physics_rigid_bodies/schema/glTF.KHR_physics_rigid_bodies.collision_filter.schema.json

spnda (approved):

PR #98, opened 2025-05-05
Material Diffuse Transmission

FishOfTheNorthStar:
Add support for KHR_material_diffuse_transmission

PR #99, opened 2025-05-05
Lots of validation fixes

nikitalita:
Hi,

I wanted to use this library for verifying and manipulating GLB assets, but I ran into a bunch of errors when attempting to validate the rewrite output. Basically, there were a bunch of bugs/missing features in serializing assets, so I have fixed these.

You'll be happy to know that we currently are able to rewrite all of the models contained in the entire gltf-sample-assets repo, and the rewritten outputs are equal to the source models.

In order to implement equality tests for the rewrite tests, I had to implement `operator==` for the GLTF objects. In addition, I made it such that it calls the `REQUIRE` catch2 macro on all the equality sub_tests if the catch2 macros are present; this is to make tracking down test failures easier. 

image

### DIFF
--- a/include/fastgltf/types.hpp
+++ b/include/fastgltf/types.hpp
@@ -2238,10 +2238,20 @@ namespace fastgltf {
 #if FASTGLTF_ENABLE_KHR_IMPLICIT_SHAPES
 	FASTGLTF_EXPORT struct SphereShape {
 		num radius = 0.5;
+
+		bool operator==(const SphereShape& other) const {
+			FASTGLTF_REQUIRE(radius == other.radius);
+			return true;
+		}
 	};
 
 	FASTGLTF_EXPORT struct BoxShape {
 		math::fvec3 size = { 1, 1, 1 };
+
+		bool operator==(const BoxShape& other) const {
+			FASTGLTF_REQUIRE(size == other.size);
+			return true;
+		}
 	};
 
 	FASTGLTF_EXPORT struct CapsuleShape {
@@ -2250,6 +2260,13 @@ namespace fastgltf {
 		num radiusBottom = 0.25;
 
 		num radiusTop = 0.25;
+
+		bool operator==(const CapsuleShape& other) const {
+			FASTGLTF_REQUIRE(height == other.height);
+			FASTGLTF_REQUIRE(radiusBottom == other.radiusBottom);
+			FASTGLTF_REQUIRE(radiusTop == other.radiusTop);
+			return true;
+		}
 	};
 
 	FASTGLTF_EXPORT struct CylinderShape {
@@ -2258,6 +2275,13 @@ namespace fastgltf {
 		num radiusBottom = 0.25;
 
 		num radiusTop = 0.25;
+
+		bool operator==(const CylinderShape& other) const {
+			FASTGLTF_REQUIRE(height == other.height);
+			FASTGLTF_REQUIRE(radiusBottom == other.radiusBottom);
+			FASTGLTF_REQUIRE(radiusTop == other.radiusTop);
+			return true;
+		}
 	};
 
 	using Shape = std::variant<SphereShape, BoxShape, CapsuleShape, CylinderShape>;
@@ -2304,6 +2328,18 @@ namespace fastgltf {
          * A multiplier applied to the acceleration due to gravity
          */
         num gravityFactor = 1;
+
+		bool operator==(const Motion& other) const {
+			FASTGLTF_REQUIRE(isKinematic == other.isKinematic);
+			FASTGLTF_REQUIRE(mass == other.mass);
+			FASTGLTF_REQUIRE(centerOfMass == other.centerOfMass);
+			FASTGLTF_REQUIRE(inertialDiagonal == other.inertialDiagonal);
+			FASTGLTF_REQUIRE(inertialOrientation == other.inertialOrientation);
+			FASTGLTF_REQUIRE(linearVelocity == other.linearVelocity);
+			FASTGLTF_REQUIRE(angularVelocity == other.angularVelocity);
+			FASTGLTF_REQUIRE(gravityFactor == other.gravityFactor);
+			return true;
+		}
 	};
 
 	FASTGLTF_EXPORT struct Geometry {
@@ -2321,6 +2357,13 @@ namespace fastgltf {
          * Flag to indicate that the geometry should be a convex hull.
          */
         bool convexHull;
+
+		bool operator==(const Geometry& other) const {
+			FASTGLTF_REQUIRE(shape == other.shape);
+			FASTGLTF_REQUIRE(node == other.node);
+			FASTGLTF_REQUIRE(convexHull == other.convexHull);
+			return true;
+		}
 	};
 
 	FASTGLTF_EXPORT struct PhysicsMaterial {
@@ -2333,6 +2376,15 @@ namespace fastgltf {
 		CombineMode frictionCombine;
 
 		CombineMode restitutionCombine;
+
+		bool operator==(const PhysicsMaterial& other) const {
+			FASTGLTF_REQUIRE(staticFriction == other.staticFriction);
+			FASTGLTF_REQUIRE(dynamicFriction == other.dynamicFriction);
+			FASTGLTF_REQUIRE(restitution == other.restitution);
+			FASTGLTF_REQUIRE(frictionCombine == other.frictionCombine);
+			FASTGLTF_REQUIRE(restitutionCombine == other.restitutionCombine);
+			return true;
+		}
 	};
 
 	FASTGLTF_EXPORT struct CollisionFilter {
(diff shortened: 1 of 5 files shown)


spnda:
Both #98 and this would now be adding `KHR_material_diffuse_transmission`. I'd also rather have this separate to a PR that "fixes validation", and silently adds an extension and fixes two extensions not being written.

Some of the test related changes I like a lot, while others seem questionable. Running an executable from within tests seems weird to me, but I understand why you're doing it. I feel like a more robust solution could be thought of there.

My main issue right now is the inconsistent formatting, and adding things here and there. For example in 49666ef7cf917653aeffac29337d21cb4eb3cfe1 the header is already included further down. You also often don't use the `std::` prefix for integer types. There's a few other issues, but I will leave those for a proper review.

However, thank you for this submission and I will tend to properly reviewing it as soon as possible. I like most of the changes. I guess the `validate` method doesn't cover everything the official glTF validator does, and therefore a lot of things weren't covered in the rewrite tests I already have.

nikitalita:
> Both https://github.com/spnda/fastgltf/pull/98 and this would now be adding KHR_material_diffuse_transmission. I'd also rather have this separate to a PR that "fixes validation", and silently adds an extension and fixes two extensions not being written. So I think it'd be best to add the GODOT and KHR extensions in a separate PR, or maybe even two.

yeah, I can do that

> Some of the test related changes I like a lot, while others seem questionable. Running an executable from within tests seems weird to me, but I understand why you're doing it. I feel like a more robust solution could be thought of there.

Unfortunately, the official validator is a Dart-only library, which can't be linked to by anything other than Dart, so the only option for calling it here is to run it as an external command 😔. 

> My main issue right now is the inconsistent formatting, and adding things here and there. For example in https://github.com/spnda/fastgltf/commit/49666ef7cf917653aeffac29337d21cb4eb3cfe1 the header is already included further down. You also often don't use the std:: prefix for integer types. There's a few other issues, but I will leave those for a proper review.

Yeah, I had noticed that; I had attempted to use Clion's formatting to compensate, but I missed some stuff. I would suggest adding a `.clang_format` to the repo in order to fix this in the future.

nikitalita:
> Both #98 and this would now be adding `KHR_material_diffuse_transmission`. I'd also rather have this separate to a PR that "fixes validation", and silently adds an extension and fixes two extensions not being written. So I think it'd be best to add the GODOT and KHR extensions in a separate PR, or maybe even two.
> 
> Some of the test related changes I like a lot, while others seem questionable. Running an executable from within tests seems weird to me, but I understand why you're doing it. I feel like a more robust solution could be thought of there.
> 
> My main issue right now is the inconsistent formatting, and adding things here and there. For example in [49666ef](https://github.com/spnda/fastgltf/commit/49666ef7cf917653aeffac29337d21cb4eb3cfe1) the header is already included further down. You also often don't use the `std::` prefix for integer types. There's a few other issues, but I will leave those for a proper review.
> 
> However, thank you for this submission and I will tend to properly reviewing it as soon as possible. I like most of the changes. I guess the `validate` method doesn't cover everything the official glTF validator does, and therefore a lot of things weren't covered in the rewrite tests I already have.

src/fastgltf.cpp
```diff
@@ -4524,6 +4560,29 @@ void fg::Exporter::writeImages(const Asset& asset, std::string& json) {
...
+		// e.g. RoughnessGrid.png -> RoughnessGrid-1.png
+		auto newPath = path;
+		int i = 1;
+		while (imagePathsHasPath(newPath)) {
+			newPath = path.parent_path() / (path.stem().string() + "-" + std::to_string(i) + path.extension().string());
+			++i;
+		}
+		imagePaths.emplace_back(newPath);
+		return newPath;
+	};
```
spnda, line 4584: Why are these lambdas? Couldn't they just be part of the `getImageFilePath` function?

tests/write_tests.cpp
```diff
@@ -102,6 +103,169 @@ TEST_CASE("Try writing a glTF with all buffers and images", "[write-tests]") {
...
+	REQUIRE(sampleFileStream.isOpen());
+	static constexpr auto requiredExtensions = static_cast<fastgltf::Extensions>(~0U);
+
+
+	//std hashset
+	std::unordered_set<std::string_view> requiredExtensionsSet;
+	for (const auto& [name, value] : fastgltf::extensionStrings) {
+		requiredExtensionsSet.insert(name);
+		REQUIRE((requiredExtensions & value) == value);
+	}
```
spnda, line 143: This is completely unused?

tests/write_tests.cpp
```diff
@@ -102,6 +103,169 @@ TEST_CASE("Try writing a glTF with all buffers and images", "[write-tests]") {
...
+	fastgltf::Parser parser(requiredExtensions);
+	auto options = fastgltf::Options::LoadExternalBuffers | fastgltf::Options::LoadExternalImages ;
+	fastgltf::Expected<fastgltf::Asset> asset = model_file.extension() == ".gltf" ?
+		parser.loadGltfJson(sampleFileStream, sampleDir, options) :
+		parser.loadGltfBinary(sampleFileStream, sampleDir, options);
+	REQUIRE(getErrorName(asset.error()) == "None");
+	REQUIRE(getErrorName(fastgltf::validate(asset.get())) == "None");
```
spnda, line 150: Why are you doing this?

spnda (changes requested):
I found quite a few major issues with the code. A lot of inconsistencies, generally bad code practices, and random comments here and there that achieve nothing, while other parts have no comments.

The biggest part, however, are the equality operators, which could've just as easily been written like this for nearly all of the types:
```cpp
bool operator==(const T& other) const = default;
```

spnda:
#95 has been merged, and I fixed a small thing, so all three of your PRs will need to be rebased

include/fastgltf/types.hpp
```diff
@@ -598,14 +608,16 @@ namespace fastgltf {
...
+        	return true;
         }
 
         // This is mostly just here for compatibility and the tests
         bool operator==(const std::vector<value_type>& other) const {
-            if (other.size() != size()) return false;
-            return std::memcmp(data(), other.data(), size_bytes()) == 0;
+            FASTGLTF_REQUIRE(other.size() == size());
+            FASTGLTF_REQUIRE(std::memcmp(data(), other.data(), size_bytes()) == 0);
+        	return true;
```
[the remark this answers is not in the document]
nikitalita: That's what was there before...? I can change it to use std::equal.

tests/write_tests.cpp
nikitalita: I wanted to get the actual error name, but since we don't return any validate errors other than "InvalidGLTF", it's pointless

nikitalita:
> The biggest part, however, are the equality operators, which could've just as easily been written like this for nearly all of the types:
> 
> ```c++
> bool operator==(const T& other) const = default;
> ```

I explicitly didn't use that because I needed to see where the tests are failing; it becomes a huge pain to debug if we can't see where the equality comparison is failing.

nikitalita:
Requested changes are done and it is rebased on top of master. I had to fix writing "KHR_physics_rigid_bodies"; we should only write it if it's actually used.

n0F4x:
> I found quite a few major issues with the code. A lot of inconsistencies, generally bad code practices, and random comments here and there that achieve nothing, while other parts have no comments.
> 
> The biggest part, however, are the equality operators, which could've just as easily been written like this for nearly all of the types:
> 
> ```c++
> bool operator==(const T& other) const = default;
> ```

This is a C++20 feature.

spnda:
> > I found quite a few major issues with the code. A lot of inconsistencies, generally bad code practices, and random comments here and there that achieve nothing, while other parts have no comments.
> > The biggest part, however, are the equality operators, which could've just as easily been written like this for nearly all of the types:
> > ```c++
> > bool operator==(const T& other) const = default;
> > ```
>
> I explicitly didn't use that because I needed to see where the tests are failing; it becomes a huge pain to debug if we can't see where the equality comparison is failing.

Yeah fair, no way of using it then.

> This is a C++20 feature.

Ah, it is indeed. Then we'll just keep it the way it is.

spnda (changes requested):
There's still a lot of style and inconsistency issues with the code in this PR, even beyond the comments in this review. Brackets, snake_case, empty lines (or the lack thereof) and random commented-out code.

I would kindly ask you to split this PR up further. I'd like to merge all of the fixes and missing functionality right away, since I think that all looks good. However, the Catch2 tests have a couple of issues with them (they don't properly work on my machine) and I'm not sure I like the setup with gltf validator yet. I think it would be best to separate these two topics, so that we can discuss the proper way of implementing this. Potentially, a bigger improvements could be made that also integrates a fuzzing mechanism as mentioned in #107. To be honest, I might even prefer if the testing functionality paired with the glTF validator is left out completely, so that I can integrate it better myself and maybe use it elsewhere in the library, too.

nikitalita:
yeah, I'm no longer interested

PR #100, opened 2025-05-05
Add support for GODOT_single_root extension

nikitalita:
This is an extension that adds no new features/objects, all it does is add validation restrictions


spnda:
> Depends on https://github.com/spnda/fastgltf/pull/99

Does it, though? I don't see why it would and I would be a lot happier if this PR would only include that commit, so that I could quickly merge this before #99, since it's a lot larger and I will need more time to review and decide. Same stands for #101.

nikitalita:
rebased on master and no longer depends on #99

#101 requires operator== changes, so it still depends on #99

spnda (approved):

PR #101, opened 2025-05-05
Add support for KHR_materials_diffuse_transmission

nikitalita:
depends on #100

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,10 @@ tests/gltf/bistro
 tests/gltf_loaders/
 tests/gltf/export/
 tests/gltf/export_glb/
+tests/gltf/export_gltf/
+tests/gltf/glTF-new-assets/
+tests/gltf/OMI_*/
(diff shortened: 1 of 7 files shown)


spnda:
> depends on https://github.com/spnda/fastgltf/pull/100

Technically, it doesn't. The `operator==` overload can be re-added in #100, after this is merged. I would prefer if this change wouldn't depend on validation improvements and new testing mechanisms.

spnda:
@FishOfTheNorthStar I would say you can re-submit a PR with either changes, from this and your old one. I had more syntax/style problems with yours iirc, and this one adds more as you said yourself. I'd say look at my review comments on your original PR, and then look at this ones and figure out what would be best. I can then give you a review on a new PR (or re-open the old one if you feel like it).

Otherwise I can probably cherry-pick the relevant stuff myself and fix it which is probably quicker but I don't really have much problem if you want to just practice all of this PR and review stuff 🤷🏻

PR #102, opened 2025-05-05
Material Diffuse Transmission

FishOfTheNorthStar:
This PR is just a follow up to an earlier one I tried but the code was getting mangled by windows cr/lf's and there were some  hurdles in getting that resolved.   So this new PR has a more correct structure.   But since I posted the original PR, someone else came along with a more full featured implementation (PR 101), so ultimately that's probably the better candidate (theirs has export support).

But I wanted to follow up anyways because it was such a mysterious thing that maybe if I explain how I resolved it, if it ever comes up in the future, the developer can refer the contributor to this archived closed PR for reference. 

To remove cr/lf's from bash:
`sed -i 's/\r$//g' ./path/to/file.cpp`

Then start a new commit, and when adding the files to it, use the --renormalize option:
`git add --renormalize ./path/to/file.cpp`

Then push the commit.  This should correct the files on Github.  

If these corrections required numerous commits, you can squash them into one with:
`git rebase -i HEAD~10`
where 10 is replaced by the number of commits you're going to remove, plus one.  In the editor that comes up, change the first commit line to replace 'pick' with 'r', and the subsequent commit lines change 'pick' to 's'.  Then pick a new commit message in the next screen, and then
`git push -f`
to push the commit with fast forward.

### DIFF
--- a/include/fastgltf/core.hpp
+++ b/include/fastgltf/core.hpp
@@ -22,7 +22,7 @@
  * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
  * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
  * OTHER DEALINGS IN THE SOFTWARE.
- */
+ */ 
 
 #ifndef FASTGLTF_CORE_HPP
 #define FASTGLTF_CORE_HPP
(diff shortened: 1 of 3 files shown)


spnda:
Interesting that you used `sed` to remove CRLF. Why didn't you just use your editor's settings to change them? But glad you got it resolved and learned a few new things about Git!

PR #105, opened 2025-05-15
Update simdjson to 3.12.3

n0F4x:


spnda:
I looked at the changelog, and the new releases since 3.11.6 do not affect the library *at all*, since it's only fixes for things we don't use or documentation changes. However, I'll still merge just for completeness.

Though I general I think updating simdjson is never really sensible, since the API is nearly completely stable and I can't see any relevant changes coming anytime soon.

PR #110, opened 2025-05-26
add mat4x4 invert function

FishOfTheNorthStar:
I was having problems using the affineInverse function to correctly invert animated cameras to their correct orientation.  It seems like there is a way to do it with affineInverse, but I have not figured out how yet.  Instead I used this function, which calculates the inverted matrix (if that's possible).  That produced the correct matrix for me for my animated cameras.

If there's a way to avoid using this function, please let me know.  I feel like I'm missing something with the affineInverse that's already in fastgltf::math... But in case it helps someone else out, here's the function that's working for me.

Credit: mat4.js from glMatrix (I just converted the code a bit)

### DIFF
--- a/include/fastgltf/math.hpp
+++ b/include/fastgltf/math.hpp
@@ -825,6 +825,65 @@ namespace fastgltf::math {
 			vec<T, 4>(l.x(), l.y(), l.z(), 1.f));
 	}
 
+	/** Inverts a 4x4 matrix, if possible. Returns true if success, false if failure.  A pointer to the input can also safely be used as the output */
+	FASTGLTF_EXPORT template <typename T>
+	bool invert(const mat<T, 4, 4>& m, mat<T, 4, 4>* output) noexcept {
+		T a00 = m[0][0];
+		T a01 = m[0][1];
+		T a02 = m[0][2];
+		T a03 = m[0][3];
+		T a10 = m[1][0];
+		T a11 = m[1][1];
+		T a12 = m[1][2];
+		T a13 = m[1][3];
+		T a20 = m[2][0];
+		T a21 = m[2][1];
+		T a22 = m[2][2];
+		T a23 = m[2][3];
+		T a30 = m[3][0];
+		T a31 = m[3][1];
+		T a32 = m[3][2];
+		T a33 = m[3][3];
+
+		T b00 = a00 * a11 - a01 * a10;
+		T b01 = a00 * a12 - a02 * a10;
+		T b02 = a00 * a13 - a03 * a10;
+		T b03 = a01 * a12 - a02 * a11;
+		T b04 = a01 * a13 - a03 * a11;
+		T b05 = a02 * a13 - a03 * a12;
+		T b06 = a20 * a31 - a21 * a30;
+		T b07 = a20 * a32 - a22 * a30;
+		T b08 = a20 * a33 - a23 * a30;
+		T b09 = a21 * a32 - a22 * a31;
+		T b10 = a21 * a33 - a23 * a31;
+		T b11 = a22 * a33 - a23 * a32;
+
+		// Calculate the determinant
+		T det = b00 * b11 - b01 * b10 + b02 * b09 + b03 * b08 - b04 * b07 + b05 * b06;
+		if (det == T(0)) {
+			// Can not invert this matrix
+			return false;
+		}
+		det = T(1) / det;
+		(*output)[0][0] = (a11 * b11 - a12 * b10 + a13 * b09) * det;
+		(*output)[0][1] = (a02 * b10 - a01 * b11 - a03 * b09) * det;
+		(*output)[0][2] = (a31 * b05 - a32 * b04 + a33 * b03) * det;
+		(*output)[0][3] = (a22 * b04 - a21 * b05 - a23 * b03) * det;
+		(*output)[1][0] = (a12 * b08 - a10 * b11 - a13 * b07) * det;
+		(*output)[1][1] = (a00 * b11 - a02 * b08 + a03 * b07) * det;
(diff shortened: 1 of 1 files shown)


include/fastgltf/math.hpp
```diff
@@ -825,6 +825,65 @@ namespace fastgltf::math {
 			vec<T, 4>(l.x(), l.y(), l.z(), 1.f));
 	}
 
+	/** Inverts a 4x4 matrix, if possible. Returns true if success, false if failure.  A pointer to the input can also safely be used as the output */
+	FASTGLTF_EXPORT template <typename T>
+	bool invert(const mat<T, 4, 4>& m, mat<T, 4, 4>* output) noexcept {
```
spnda, line 830: why is the output a pointer and not a reference?

include/fastgltf/math.hpp
```diff
@@ -825,6 +825,65 @@ namespace fastgltf::math {
...
+		T b09 = a21 * a32 - a22 * a31;
+		T b10 = a21 * a33 - a23 * a31;
+		T b11 = a22 * a33 - a23 * a32;
+
+		// Calculate the determinant
+		T det = b00 * b11 - b01 * b10 + b02 * b09 + b03 * b08 - b04 * b07 + b05 * b06;
+		if (det == T(0)) {
+			// Can not invert this matrix
+			return false;
+		}
```
spnda, line 866: While I agree with this, the other inversion functions don't check this either. I know it's not *great*, but for example `glm` assumes the same, and always just returns the new matrix, and reduces some code around the caller. This also removes the second argument rendering my last comment moot.

include/fastgltf/math.hpp
```diff
@@ -825,6 +825,65 @@ namespace fastgltf::math {
...
+		det = T(1) / det;
```
spnda, line 867: While yes, mul ops are generally faster than div, I'd prefer if this operator would be applied to the entire matrix at the end, making the code a little simpler to read. I guess the `mat::operator/` should maybe make use of this minor optimization.
You would need to change the following function, and you could include it in this PR if you want, however you need to be careful since this trick only works for floating point numbers! So I guess you'd need to use SFINAE or concepts to differentiate between the two cases.
https://github.com/spnda/fastgltf/blob/159eb795d25c4466d14d209942959326007a0766/include/fastgltf/math.hpp#L747-L752

include/fastgltf/math.hpp
```diff
@@ -825,6 +825,65 @@ namespace fastgltf::math {
...
+		T b05 = a02 * a13 - a03 * a12;
+		T b06 = a20 * a31 - a21 * a30;
+		T b07 = a20 * a32 - a22 * a30;
+		T b08 = a20 * a33 - a23 * a30;
+		T b09 = a21 * a32 - a22 * a31;
+		T b10 = a21 * a33 - a23 * a31;
+		T b11 = a22 * a33 - a23 * a32;
+
+		// Calculate the determinant
+		T det = b00 * b11 - b01 * b10 + b02 * b09 + b03 * b08 - b04 * b07 + b05 * b06;
```
spnda, line 862: There's a `determinant` function for both 2x2 and 3x3 matrices, but not for 4x4. I think it would be better if this would be moved to a dedicated function, so that it can be used directly and we don't duplicate code.

spnda (changes requested):
I find it fairly obvious that this code was copied from glm or something like that, and I'd prefer if the code style aligned with the rest of the header.

Also, regarding your issue, I think camera transformations are simply not affine, which could be the cause for your issue but I am not an expert in linear algebra so I can't say for sure.

include/fastgltf/math.hpp
FishOfTheNorthStar: That seems like it could create additional overhead / redundant operations when the invert() function gets it's own references into source matrix and calculates the b__ values, since it uses both the a__ and b__ set's after it checks the determinant.   We could put a copy of the determinant code into it's own function for isolated use, that sounds useful.  But I don't think it would be good to actually use it within invert() since it needs so much of the same precursor variables.

include/fastgltf/math.hpp
FishOfTheNorthStar: Yeah, once I got my source matrix formed correctly, this never occurred.  At first I wasn't scaling the 3x3 rot part of the 4x4 correctly and this situation did occur, so I had testing in place for it that I thought might be helpful.  Like you say it's probably safe to remove this test, if someone is supplying a bad source matrix, no amount of testing will make it valid, so just return identity?  Then if users really want to test if this happened, they could check if the result was exactly identity, which should never be the case in successful operation.

include/fastgltf/math.hpp
FishOfTheNorthStar: I added a mat4x4 determinant function ( but it's not used within the invert function )

include/fastgltf/math.hpp
FishOfTheNorthStar: true/false return value removed, pointer/reference to output removed, now it just outputs a new mat4x4 with the inverted value or identity if the determinant was 0

include/fastgltf/math.hpp
FishOfTheNorthStar: parameter removed

include/fastgltf/math.hpp
spnda: It's fairly limited as an optimization, especially on modern processors. I just recently looked into assembly for a microprocessor I was using and I think a single div took about 4 times as long at around 20 cycles, but this is a fairly old arm microprocessor so it's not all to comparable to modern x86 or armv8 chips

include/fastgltf/math.hpp
spnda: Yeah I think that's fair

PR #114, opened 2025-07-01
fix: redudant moves and unused variables warnings

AndreCostaaa:
Hi fantastic project you have here

While integrating it into a project with `-Wall -Wextra -Wpedantic` I got a bunch of warnings that I figured would resolve and contribute upstream

Let me know what you think

To reproduce the warnings:

```bash
❯ cmake -B build -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wpedantic"
-- The C compiler identification is GNU 14.3.1
-- The CXX compiler identification is GNU 14.3.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Could NOT find simdjson (missing: simdjson_DIR)
-- fastgltf: Found simdjson (Version 3.12.3)
-- Configuring done (0.4s)
-- Generating done (0.0s)
-- Build files have been written to: /home/andre/dev/lvgl/fastgltf/build
❯ cmake --build build -j$(nproc)
[ 40%] Building CXX object CMakeFiles/fastgltf.dir/src/fastgltf.cpp.o
[ 40%] Building CXX object CMakeFiles/fastgltf.dir/deps/simdjson/simdjson.cpp.o
[ 80%] Building CXX object CMakeFiles/fastgltf.dir/src/base64.cpp.o
[ 80%] Building CXX object CMakeFiles/fastgltf.dir/src/io.cpp.o
In file included from /home/andre/dev/lvgl/fastgltf/src/fastgltf.cpp:51:
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp: In static member function ‘static fastgltf::Expected<fastgltf::GltfDataBuffer> fastgltf::GltfDataBuffer::FromPath(const std::filesystem::__cxx11::path&)’:
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp:628:41: warning: redundant move in return statement [-Wredundant-move]
  628 |                         return std::move(buffer);
      |                                ~~~~~~~~~^~~~~~~~
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp:628:41: note: remove ‘std::move’ call
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp: In static member function ‘static fastgltf::Expected<fastgltf::GltfDataBuffer> fastgltf::GltfDataBuffer::FromBytes(const std::byte*, std::size_t)’:
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp:636:41: warning: redundant move in return statement [-Wredundant-move]
  636 |                         return std::move(buffer);
      |                                ~~~~~~~~~^~~~~~~~
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp:636:41: note: remove ‘std::move’ call
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp: In static member function ‘static fastgltf::Expected<fastgltf::MappedGltfFile> fastgltf::MappedGltfFile::FromPath(const std::filesystem::__cxx11::path&)’:
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp:703:41: warning: redundant move in return statement [-Wredundant-move]
  703 |                         return std::move(buffer);
      |                                ~~~~~~~~~^~~~~~~~
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp:703:41: note: remove ‘std::move’ call
In file included from /home/andre/dev/lvgl/fastgltf/src/io.cpp:29:
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp: In static member function ‘static fastgltf::Expected<fastgltf::GltfDataBuffer> fastgltf::GltfDataBuffer::FromPath(const std::filesystem::__cxx11::path&)’:
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp:628:41: warning: redundant move in return statement [-Wredundant-move]
  628 |                         return std::move(buffer);
      |                                ~~~~~~~~~^~~~~~~~
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp:628:41: note: remove ‘std::move’ call
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp: In static member function ‘static fastgltf::Expected<fastgltf::GltfDataBuffer> fastgltf::GltfDataBuffer::FromBytes(const std::byte*, std::size_t)’:
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp:636:41: warning: redundant move in return statement [-Wredundant-move]
  636 |                         return std::move(buffer);
      |                                ~~~~~~~~~^~~~~~~~
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp:636:41: note: remove ‘std::move’ call
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp: In static member function ‘static fastgltf::Expected<fastgltf::MappedGltfFile> fastgltf::MappedGltfFile::FromPath(const std::filesystem::__cxx11::path&)’:
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp:703:41: warning: redundant move in return statement [-Wredundant-move]
  703 |                         return std::move(buffer);
      |                                ~~~~~~~~~^~~~~~~~
/home/andre/dev/lvgl/fastgltf/include/fastgltf/core.hpp:703:41: note: remove ‘std::move’ call
/home/andre/dev/lvgl/fastgltf/src/io.cpp: In member function ‘virtual fastgltf::span<std::byte> fastgltf::GltfDataBuffer::read(std::size_t, std::size_t)’:
/home/andre/dev/lvgl/fastgltf/src/io.cpp:103:77: warning: unused parameter ‘padding’ [-Wunused-parameter]
  103 | fg::span<std::byte> fg::GltfDataBuffer::read(std::size_t count, std::size_t padding) {
      |                                                                 ~~~~~~~~~~~~^~~~~~~
/home/andre/dev/lvgl/fastgltf/src/io.cpp: In member function ‘virtual fastgltf::span<std::byte> fastgltf::MappedGltfFile::read(std::size_t, std::size_t)’:
/home/andre/dev/lvgl/fastgltf/src/io.cpp:297:77: warning: unused parameter ‘padding’ [-Wunused-parameter]
  297 | fg::span<std::byte> fg::MappedGltfFile::read(std::size_t count, std::size_t padding) {
      |                                                                 ~~~~~~~~~~~~^~~~~~~
/home/andre/dev/lvgl/fastgltf/src/fastgltf.cpp: In member function ‘fastgltf::Expected<fastgltf::Asset> fastgltf::Parser::parse(simdjson::dom::object, fastgltf::Category)’:
/home/andre/dev/lvgl/fastgltf/src/fastgltf.cpp:1621:25: warning: redundant move in return statement [-Wredundant-move]
 1621 |         return std::move(asset);
      |                ~~~~~~~~~^~~~~~~
/home/andre/dev/lvgl/fastgltf/src/fastgltf.cpp:1621:25: note: remove ‘std::move’ call
/home/andre/dev/lvgl/fastgltf/src/fastgltf.cpp: In member function ‘fastgltf::Expected<fastgltf::ExportResult<std::__cxx11::basic_string<char> > > fastgltf::Exporter::writeGltfJson(const fastgltf::Asset&, fastgltf::ExportOptions)’:
/home/andre/dev/lvgl/fastgltf/src/fastgltf.cpp:7039:21: warning: redundant move in return statement [-Wredundant-move]
 7039 |     return std::move(result);
      |            ~~~~~~~~~^~~~~~~~
/home/andre/dev/lvgl/fastgltf/src/fastgltf.cpp:7039:21: note: remove ‘std::move’ call
/home/andre/dev/lvgl/fastgltf/src/fastgltf.cpp: In member function ‘fastgltf::Expected<fastgltf::ExportResult<std::vector<std::byte, std::allocator<std::byte> > > > fastgltf::Exporter::writeGltfBinary(const fastgltf::Asset&, fastgltf::ExportOptions)’:
/home/andre/dev/lvgl/fastgltf/src/fastgltf.cpp:7136:21: warning: redundant move in return statement [-Wredundant-move]
 7136 |     return std::move(result);
      |            ~~~~~~~~~^~~~~~~~
/home/andre/dev/lvgl/fastgltf/src/fastgltf.cpp:7136:21: note: remove ‘std::move’ call
[100%] Linking CXX static library libfastgltf.a
[100%] Built target fastgltf

### DIFF
--- a/include/fastgltf/core.hpp
+++ b/include/fastgltf/core.hpp
@@ -625,7 +625,7 @@ namespace fastgltf {
 			if (buffer.error != fastgltf::Error::None) {
 				return buffer.error;
 			}
-			return std::move(buffer);
+			return buffer;
 		}
 
 		static Expected<GltfDataBuffer> FromBytes(const std::byte* bytes, std::size_t count) noexcept {
@@ -633,7 +633,7 @@ namespace fastgltf {
 			if (buffer.error != fastgltf::Error::None) {
 				return buffer.error;
 			}
-			return std::move(buffer);
+			return buffer;
 		}
 
 #if FASTGLTF_CPP_20
@@ -642,7 +642,7 @@ namespace fastgltf {
 			if (buffer.buffer.get() == nullptr) {
 				return buffer.error;
 			}
-			return std::move(buffer);
+			return buffer;
 		}
 #endif
 
@@ -700,7 +700,7 @@ namespace fastgltf {
 			if (buffer.error != fastgltf::Error::None) {
 				return buffer.error;
 			}
-			return std::move(buffer);
+			return buffer;
 		}
 
 		void read(void* ptr, std::size_t count) override;
@@ -761,7 +761,7 @@ namespace fastgltf {
 			if (buffer.buffer.get() == nullptr) {
 				return buffer.error;
 			}
-			return std::move(buffer);
+			return buffer;
 		}
 	};
 	#endif
--- a/src/io.cpp
+++ b/src/io.cpp
@@ -100,7 +100,7 @@ void fg::GltfDataBuffer::read(void *ptr, std::size_t count) {
 	idx += count;
 }
 
-fg::span<std::byte> fg::GltfDataBuffer::read(std::size_t count, std::size_t padding) {
+fg::span<std::byte> fg::GltfDataBuffer::read(std::size_t count, [[maybe_unused]] std::size_t padding) {
 	span<std::byte> sub(buffer.get() + idx, count);
 	idx += count;
 	return sub;
@@ -294,7 +294,7 @@ void fg::MappedGltfFile::read(void *ptr, std::size_t count) {
 	idx += count;
 }
 
-fg::span<std::byte> fg::MappedGltfFile::read(std::size_t count, std::size_t padding) {
+fg::span<std::byte> fg::MappedGltfFile::read(std::size_t count, [[maybe_unused]] std::size_t padding) {
 	span<std::byte> sub(static_cast<std::byte*>(mappedFile) + idx, count);
 	idx += count;
 	return sub;
--- a/src/fastgltf.cpp
+++ b/src/fastgltf.cpp
@@ -1618,7 +1618,7 @@ fg::Expected<fg::Asset> fg::Parser::parse(simdjson::dom::object root, Category c
 		}
 	}
 
-	return std::move(asset);
+	return asset;
 }
 
 fg::Error fg::Parser::parseAccessors(simdjson::dom::array& accessors, Asset& asset) {
@@ -7036,7 +7036,7 @@ fg::Expected<fg::ExportResult<std::string>> fg::Exporter::writeGltfJson(const As
     result.output = std::move(outputString);
     result.bufferPaths = std::move(bufferPaths);
     result.imagePaths = std::move(imagePaths);
-    return std::move(result);
+    return result;
 }
 
 fg::Expected<fg::ExportResult<std::vector<std::byte>>> fg::Exporter::writeGltfBinary(const Asset& asset, ExportOptions _options) {
@@ -7133,7 +7133,7 @@ fg::Expected<fg::ExportResult<std::vector<std::byte>>> fg::Exporter::writeGltfBi
         }
     }
 
-    return std::move(result);
+    return result;
 }
 
 namespace fastgltf {

### CODE REVIEW

**Comment by AndreCostaaa** [comment]
Hi @spnda 

I was a bit confused by what you meant and only then realized that I had actually removed `std::move` from places where it definitely makes sense to keep them. Just fixed this PR

Could I bother to take a look at this again ? You will find that the actual `std::move` that should be removed and were causing warnings were from functions that try to move while returning local variables which actually prevents the compiler from performing some optimizations like constructing the object in the caller's memory location directly and it's why the warning arises

Sorry for the confusion

**Comment by spnda** [comment]
> Could I bother to take a look at this again ? You will find that the actual `std::move` that should be removed and were causing warnings were from functions that try to move while returning local variables which actually prevents the compiler from performing some optimizations like constructing the object in the caller's memory location directly and it's why the warning arises

Yes, `return std::move(foo)` is usually bad practice since it will make the compiler unable to perform RVO or NRVO. However, that's not what's happening here. These functions return `Expected<T>` but the type being moved is `T`z What's actually happening here is we're constructing a `Expected<T>` using the `Expected(T&&)` constructor, and then relying on the compiler to perform RVO on that object. You can check this by specifying `-Wnrvo` (iirc that's the name), which will give a warning when the compiler is unable to perform RVO, and that shouldn't be the case here. Please correct me if I'm wrong, though.

**Comment by AndreCostaaa** [comment]
The changes to `Expected<T>` are the ones that weren't actually producing warnings and I had removed them by mistake. I forced-push this branch to revert those changes

File: src/io.cpp [hunk 1]
```diff
@@ -295,6 +296,7 @@ void fg::MappedGltfFile::read(void *ptr, std::size_t count) {
 }
 
 fg::span<std::byte> fg::MappedGltfFile::read(std::size_t count, std::size_t padding) {
+	(void)padding;
```
spnda, line 299: This is C code, and looks horrible. C++ has a `[[maybe_unused]]` attribute that should be used in these cases.

src/io.cpp
```diff
@@ -101,6 +101,7 @@ void fg::GltfDataBuffer::read(void *ptr, std::size_t count) {
 }
 
 fg::span<std::byte> fg::GltfDataBuffer::read(std::size_t count, std::size_t padding) {
+	(void)padding;
```
spnda, line 104: Same `[[maybe_unused]]`.

spnda (changes requested):

include/fastgltf/core.hpp
```diff
@@ -625,15 +625,15 @@ namespace fastgltf {
 			if (buffer.error != fastgltf::Error::None) {
 				return buffer.error;
 			}
-			return std::move(buffer);
```
AndreCostaaa, line 628: I don't understand what you mean, this is returning a local variable and not some templated type. Am I missing something?

src/io.cpp
AndreCostaaa: Agreed that it doesn't belong in a C++ codebase but wanted to know how you wanted to deal with it exactly

AndreCostaaa:
For member templated files I completely agree!

What I'm modifying is cases where we return a local variable and in that case it doesn't make sense to be explicit about the move since we would prevent the compiler from performing RVO (thank you for pointing me to the correct term)

include/fastgltf/core.hpp
spnda: This is not returning a local variable, as I tried to explain earlier.

There is an implicit constructor being called that constructs an `Expected<GltfDataBuffer>` object from a `GltfDataBuffer&&`, which is what the `std::move(buffer)` returns.

You can think of it more like this:
```cpp
GltfDataBuffer buffer;
Expected<GltfDataBuffer> expected(std::move(buffer));
return expected;
```

spnda: Yes, I understand this and you obviously do too. However, your premise is wrong:
> Buffer is a local variable being returned

`buffer` is not returned. That's the misconception here I'm trying to address. See my comment above of what the code expands to effectively due to the implicit constructor. We're returning an implicitly created object, which is *not* moved. It is only created using a moved object of `GltfDataBuffer`, hence the `std::move(buffer)`.

As a simple example, the following code compiled using `gcc -std=c++23 -Wnrvo` does not throw *any* warnings, meaning that the copies for returned objects was elided:
```cpp
enum class error { err, };
std::expected<std::string, error> foo() {
  std::string value;
  return std::move(value);
}
```
See https://godbolt.org/z/eT5KhzxEr

src/io.cpp
spnda:
```cpp
fg::span<std::byte> fg::MappedGltfFile::read(std::size_t count, [[maybe_unused]] std::size_t padding) {
```
That's about the best you can get with C+11 and newer. C++26 has placeholders I think so `std::size_t _` is also a thing.

include/fastgltf/core.hpp
AndreCostaaa: No worries, it's definitely interesting to dive into how these details work. Thank you!

src/io.cpp
spnda: > is a thing in cpp as I've seen it used before

meh, it's mostly a legacy and old thing to do, but yes a lot of people do that with various keywords or variable names, but since C++11 gives the attribute there's no need to do this in a modern codebase.

spnda (changes requested):
Given that you've changed all standard factory constructors of `GltfDataBuffer`, you can also change the one guarded by `FASTGLTF_CPP_20`, which would have the same issue. Also, since we've established the warnings to be correct, what about the templated functions? Could those be re-added?

AndreCostaaa:
> Given that you've changed all standard factory constructors of GltfDataBuffer, you can also change the one guarded by FASTGLTF_CPP_20, which would have the same issue.

Done, there's also one guarded by `__ANDROID__` that should have the same problem but since I can't test i would rather not touch it

> Also, since we've established the warnings to be correct, what about the templated functions? Could those be re-added?

The only `std::move` left in `core.hpp` (besides the one guarded by `__ANDROID__`) is this one:

```cpp
		[[nodiscard]] T&& operator*() && noexcept {
			assert(err == Error::None);
			return std::move(value);
		}
```
Which is moving a member variable so I believe this is necessary as the default here would be to copy to not invalidate a member variable

There's some in `types.hpp` but also return member variables like this one so I also believe it's correct to use `std::move` in those cases

AndreCostaaa:
Thank you very much :rocket:

spnda:
@AndreCostaaa May I ask if this is for LVGL or for one of your own projects? I'm currently doing embedded stuff with Zephyr so it's nice to see this project making its way into embedded environments, too, if that's the case.

PR #115, opened 2025-07-03
Missing <bit> for std::byteswap in util.hpp

stripe2933:
```
vcpkg_installed/x64-linux/include/fastgltf/util.hpp:392:15: error: missing '#include <bit>'; 'byteswap' must be declared before it is used
  392 |                 return std::byteswap(n);
      |                             ^
/usr/lib/gcc/x86_64-linux-gnu/15/../../../../include/c++/15/bit:112:5: note: declaration here is not visible
  112 |     byteswap(_Tp __value) noexcept
      |     ^
1 error generated.
```

Fix compilation error in GCC-15.

### DIFF
--- a/include/fastgltf/util.hpp
+++ b/include/fastgltf/util.hpp
@@ -29,6 +29,7 @@
 
 #if !defined(FASTGLTF_USE_STD_MODULE) || !FASTGLTF_USE_STD_MODULE
 #include <array>
+#include <bit>
 #include <cmath>
 #include <cstddef>
 #include <cstdint>


spnda:
1. `<bit>` is a C++20 header and can't be included without guards
2. It is already included with guards here: https://github.com/spnda/fastgltf/blob/4cf17068c3b2b0ae89128c5f4ae03d5891aa4885/include/fastgltf/util.hpp#L71

The issue seems like it stems from something else, and not a missing include. Perhaps `__cpp_lib_bitops` isn't defined or has a different value? Thinking about this for just a few seconds would've been sufficient for noticing the very obvious issues this PR has.

PR #120, opened 2025-10-04
build: check for file download status when downloading simdjson

AndreCostaaa:
When CMake's `file(DOWNLOAD)` command fails, it silently creates empty files at the destination without reporting any errors. The build continues successfully through the CMake configuration phase, but then fails during compilation with confusing errors about missing declarations or undefined symbols. This makes it extremely difficult to diagnose that the root cause was actually a failed download, since the empty files exist and there's no indication that the download step failed.

This PR adds explicit status checking after each `file(DOWNLOAD)` which immediately catches download failures during the CMake configuration phase


AndreCostaaa:
I'm confused why the CI is failing for this :thinking:

spnda:
the fails are unrelated to your changes and are happening on master, too. I have yet to check for why this is, but it seems that the sample assets have changed somewhere making my tests fail, since they assume a lot of specific values for the assets.

PR #123, opened 2025-11-17
fix: correct order of operations for TRS encoded transforms

FishOfTheNorthStar:
the existing way looks like (in tools.hpp):
`return scale(rotate(translate(base, trs.translation), trs.rotation), trs.scale);`

This applies the translation first since it's inner-most, then the rotation, then the scaling.  But the glTF standard calls for the opposite of that, it should be scale first, then rotation, then translation.

> TRS properties are converted to matrices and **postmultiplied** in the T * R * S order to compose the transformation matrix; first the scale is applied to the vertices, then the rotation, and then the translation.

[https://kcoley.github.io/glTF/specification/2.0/](https://kcoley.github.io/glTF/specification/2.0/#transformations)

So this appears to be backwards to me.  It probably doesn't come up a whole lot since the transforms are usually stored as 4x4 matrices, but I noticed the discrepancy while working on other stuff here so I thought I'd pass it along.

edit: for addtional reference on this subject see link below
[https://github.com/javagl/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_004_ScenesNodes.md#local-transforms-of-nodes](https://github.com/javagl/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_004_ScenesNodes.md#local-transforms-of-nodes)

### DIFF
--- a/include/fastgltf/tools.hpp
+++ b/include/fastgltf/tools.hpp
@@ -901,7 +901,7 @@ FASTGLTF_EXPORT inline auto getTransformMatrix(const Node& node, const math::fma
 			return base * matrix;
 		},
 		[&](const TRS& trs) {
-			return scale(rotate(translate(base, trs.translation), trs.rotation), trs.scale);
+			return translate(rotate(scale(base, trs.scale), trs.rotation), trs.translation);
 		}
 	}, node.transform);
 }


FishOfTheNorthStar:
> > TRS properties are converted to matrices and postmultiplied in the T * R * S order to compose the transformation matrix; first the scale is applied to the vertices, then the rotation, and then the translation.
> 
> This applies to the vertices, and it does not directly describe the order in which to calculate the matrix.
> 
> I'm not a 100% sure anymore since it's been a while since I've done matrix stuff, but to me this proposal sounds wrong. If you were to take the matrix and apply scale first, and then translate, your resulting scale is wrong. That's why, at least in the matrix calculation, we translate first and then scale. The link you gave from the glTF-Tutorial also agrees with this and the current code. So I'm fairly sure the current implementation is correct.

the pseudo code may be a bit misleading, the line where you see `localTransform = translationMatrix * rotationMatrix * scaleMatrix;` happens from right to left, so first is rot times scale, then the product of that is multiplied by the translation matrix at the end.

but the part i'm unclear about now is maybe even though the order of processing is, in current form, backwards from that... maybe it's still actually applied in a 'post-multiplied' (ie, right to left) way?

I've turned to the AI to see if it has any thoughts on this.  It seems to agree with me, more than I agree with me actually.  I still like the current way better even if it isn't technically correct (but it might be, i could be misreading all of this).

image
image

FishOfTheNorthStar:
hey great.  yeah that's a solid wall of rant there and spoiler: half of it is flat out wrong at the end, lol (yep, it's late).  I think that's a good call though man, thanks for hearing me out.

I'm terrible at git, you may recall that from a while ago and it has not improved much.  If you need anything from me to do the cherrypick thing let me know.

FishOfTheNorthStar:
**REVERT IMMEDIATELY I WAS WRONG - REVERT IMMEDIATELY I WAS WRONG.**

So... about this.  I think I was wrong.  Basically you were doing it right and everything on this PR is based on the assumption you were composing unique matrices for the T/R/S components within those translate/rotate/scale functions, and then multiplying that matrix result times the input matrix.  If that's what you were doing, all of this is accurate... but it's not what you're doing.  Not at all.  The translate/rotate/scale functions are operating 'in place', they aren't creating a new matrix and multiplying it against the input, they're modifying the input matrix directly.  Because that is the case, you need the order to be 'backwards' like this appears to be.


Example (with structure from before this PR):
```
decomposeTransformMatrix( old_matrix, local_scale, local_quat, local_pos);
new_matrix = scale( rotate( translate ( fmat4x4(), local_pos),  local_quat), local_scale);
```

new matrix will equal old matrix, even though the scale is applied last.
if you reverse the order (ie, use this PR), new matrix will not even remotely equal old matrix.

REVERT IMMEDIATELY!  omg man.  I'm so sorry, it really looked weird to me, and it for sure is, but it was weird in a way I didn't fully understand, and I had to unpack it a while.  

You could keep that getLocalTransformMatrix part in there though, that's very helpful.

Further explanation, what went wrong here:
Let's say you did it instead like this:
```
decomposeTransformMatrix( old_matrix, local_scale, local_quat, local_pos);
scale_matrix = scale( fmat4x4(), local_scale);
rotate_matrix = rotate( fmat4x4(), local_quat);
translation_matrix = translate ( fmat4x4(), local_pos);

```
Now in this case, everything I was saying would be right, if you created new matrix like this:
`new_matrix = translation_matrix * rotate_matrix * scale_matrix`
then in that case, new_matrix will equal old_matrix.  That is what I thought was happening internally.  I was wrong about that.

conversly, if you created it like this:
`new_matrix = scale * rotate * translate`, ie translate first, it would be broken.  This is how it appeared to be doing it to me, but I've been testing it against different scenarios here today a few different ways, and I finally just started taking transforms that I knew were good and fine and then decomposing them and reforming them, and making sure you end up with the same thing you started with.  You did before this PR, but you don't now, with it, so it's just... completely broken.

Man.  I feel awful.  I just now got a situation put together to test this thoroughly, and when i got confirmation it's not right,  I came here right away to ask you to reconsider merging this PR.  

Not only does the way your'e doing it inexplicably work despite seeming like it shouldn't, it's probably a lot faster than doing it one matrix per stage and doing the full matrix mult for each stage.  So even if you wanted it to read like the standard says, and be more 'clear' like that, you probably still shouldn't because it will run slower.

I need to look through your translate/rotate/scale functions more to understand them better.  Maybe once I've had a while to really unpack it all and see if there's anything that doesn't check out, I can revisit this with those findings.  But until then, REVERT REVERT REVERT.

FishOfTheNorthStar:
I'm starting to think even the https://github.com/spnda/fastgltf/commit/eccd8ccd2ab698844b5a08d8368fc05b713e6f0c refactor might be a bad idea.  Because I'm going through and counting math operations, either way, and if you figure one matrix multiplication is 64 mults, but the 'modify in-place' approach you were taking before is a lot less.... that's just a lot of math overhead to add for the sake of code clarity.

Yeah i'd just roll back all of this actually.  I meant well, sorry to rope you into a bad plan here like this.   The getLocalTransfromMatrix is nice on it's own but it probably shouldn't be required within getTransformMatrix, that will slow things down and provide no benefit.

spnda:
> Man. I feel awful. 

It's all good man, I had no idea either what was wrong or right lol. I will just quickly revert the PR myself, I don't think the additional comment in #127 is really necessary. I've just reverted it here: d3d6ee6.

Thank you anyway.

PR #127, opened 2025-11-20
Revert PR#123

FishOfTheNorthStar:
My previous PR was wrong and should be removed.  I added a short note to the line in question to help clarify.  In this PR, the 'getLocalTransformMatrix' function is still added, but the order of operations is restored to the per-PR123 state.

Sorry for the hassle.  Math is hard.


spnda:
See comment https://github.com/spnda/fastgltf/pull/123#issuecomment-3591540503. I reverted it myself. Thank you, anyway.

Issue #133, opened 2026-01-09
Mixed tabs and spaces used in headers


slowriot:
There's currently a mix of tabs and spaces used for indentation in the headers (sometimes both within a single function or line).  It mostly looks alright if you have your tabs set to render as four spaces, but with any other setting the result is a mess.

Unfortunately it looks like it's about 50/50 tabs vs spaces across the board, so a fairly large change will be required to make it entirely consistent.

spnda:
I have just gone through everything now and forcibly changed it in 67942b465cafa2ea7644654a4eb64138ff038a2c. I will again think about an editor config or something else soon...

PR #134, opened 2026-01-09
Fix some mixed tabs and spaces in headers

slowriot:
This addresses part of, but does not fully resolve, issue #133.

There's currently a mix of tabs and spaces used for indentation in the headers (sometimes both within a single function or line). It mostly looks alright if you have your tabs set to render as four spaces, but with any other setting the result is a mess.

Unfortunately it looks like it's about 50/50 tabs vs spaces across the board, so a fairly large change will be required to make it entirely consistent.

This PR just addresses those parts of the headers where tabs and spaces were used in the same place (mixed within one function, or multiple types on the same line).  It leaves alone the rest of the file where there are consistent sections of spaces or tabs used together.

Only whitespace is changed in this PR.


spnda:
I've just gone ahead and converted the entire project to use Tabs: 67942b465cafa2ea7644654a4eb64138ff038a2c

PR #147, opened 2026-07-11
Exporter: fix KHR_lights_punctual spot light serialization

tksuoran:
Exporter::writeLights wrote innerConeAngle / outerConeAngle at the light's top level with broken comma placement: no separator after "type" (or "range") and a trailing comma before "name" / the closing brace, so any spot light with cone angles produced invalid JSON.

Per the KHR_lights_punctual schema the cone angles belong inside a nested "spot" object -- which is also where this library's own parser reads them (parseLightsPunctual), so exported spot lights did not round-trip even ignoring the JSON syntax errors.

Write ',"spot":{"innerConeAngle":...,"outerConeAngle":...}' instead.


spnda (approved):
Seems fair

PR #148, opened 2026-07-23
Fix SmallVector move constructor and move assignment for move-only types

jihaohaaaa:
When FASTGLTF_USE_CUSTOM_SMALLVECTOR is enabled, SmallVector(SmallVector&&) and SmallVector::operator=(SmallVector&&) previously invoked copy(), which uses copy assignment *result++ = *first. For types with deleted copy constructors/assignments (such as Primitive containing std::unique_ptr<DracoCompressedPrimitive>), this caused compilation errors when moving SmallVector<Primitive>. This PR introduces a move_elements helper function to perform std::move instead of copying.

### DIFF
--- a/include/fastgltf/types.hpp
+++ b/include/fastgltf/types.hpp
@@ -730,14 +730,21 @@ namespace fastgltf {
 		std::size_t _size = 0, _capacity = N;
 
 		void copy(const T* first, std::size_t count, T* result) {
-			if (count > 0) {
-				if constexpr (std::is_trivially_copyable_v<T>) {
-					std::memcpy(result, first, count * sizeof(T));
-				} else {
-					*result++ = *first;
-					for (std::size_t i = 1; i < count; ++i) {
-						*result++ = *++first;
-					}
+			if constexpr (std::is_trivially_copyable_v<T>) {
+				std::memcpy(result, first, count * sizeof(T));
+			} else {
+				for (std::size_t i = 0; i < count; ++i) {
+					result[i] = first[i];
+				}
+			}
+		}
+
+		void move_elements(T* first, std::size_t count, T* result) {
+			if constexpr (std::is_trivially_copyable_v<T>) {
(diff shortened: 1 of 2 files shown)


include/fastgltf/types.hpp
```diff
@@ -729,15 +729,29 @@ namespace fastgltf {
         std::size_t _size = 0, _capacity = N;
 
         void copy(const T* first, std::size_t count, T* result) {
+            if constexpr (std::is_copy_assignable_v<T> || std::is_copy_constructible_v<T>) {
```
spnda: Wouldn't it be better to omit this to get compile errors in case the copy() branch is called erroneously instead of silently not copying?

include/fastgltf/types.hpp
```diff
@@ -729,15 +729,29 @@ namespace fastgltf {
...
+                if (count > 0) {
+                    if constexpr (std::is_trivially_copyable_v<T>) {
+                        std::memcpy(result, first, count * sizeof(T));
+                    } else {
+                        *result++ = *first;
+                        for (std::size_t i = 1; i < count; ++i) {
+                            *result++ = *++first;
+                        }
+                    }
+                }
```
spnda: I don't remember why I wrote it like this and why the `count > 0` check, but couldn't this also be simplified while we're at it?

include/fastgltf/types.hpp
```diff
@@ -729,15 +729,29 @@ namespace fastgltf {
...
+            }
+        }
+
+        void move_elements(T* first, std::size_t count, T* result) {
             if (count > 0) {
```
spnda: same thing about the check

spnda (changes requested):

jihaohaaaa:
Updated. Rebased on latest main, simplified `copy`/`move_elements` and removed the `if constexpr`, and added additional tests for move-only types.

spnda (approved):

spnda:
Thanks!

PR #149, opened 2026-08-04
add new mesh field

### LINKED ISSUE #142: Support KHR_physics_rigid_bodies
Newer draft of what used to be KHR_physics_rigid_body has a bunch of changes.

Reference: https://github.com/eoineoineoin/glTF_Physics

DethRaid:
this pr changes the `node` field on the `Geometry` object of `KHR_physics_rigidbodies` to be `mesh`. this bring this library in line with the current version of that extension

this does not add the plane shape because my game does not use it. and  it does not change the fastgltf version number even though this was a braking change

partially resolves #142

### DIFF
--- a/include/fastgltf/types.hpp
+++ b/include/fastgltf/types.hpp
@@ -36,16 +36,14 @@
 #include <vector>
 #endif
 
-// Utils header already includes some headers, which we'll try and avoid
-// including twice.
-#include <fastgltf/math.hpp>
+// Utils header already includes some headers, which we'll try and avoid including twice.
 #include <fastgltf/util.hpp>
+#include <fastgltf/math.hpp>
 
 #if defined(_GLIBCXX_USE_CXX11_ABI) && !_GLIBCXX_USE_CXX11_ABI
 // polymorphic allocators are only supported with the 'new' GCC ABI.
-// Older compilers (older than GCC 5.1) default to the old ABI and sometimes the
-// old ABI is explicitly selected on even the newest compilers, which we want to
-// support.
+// Older compilers (older than GCC 5.1) default to the old ABI and sometimes the old ABI is
+// explicitly selected on even the newest compilers, which we want to support.
 #define FASTGLTF_DISABLE_CUSTOM_MEMORY_POOL 1
 #endif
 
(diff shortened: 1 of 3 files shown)


src/fastgltf.cpp
```diff
@@ -4688,8 +4688,8 @@ fg::Error fg::Parser::parsePhysicsRigidBody(simdjson::dom::object& khr_physics_r
 				return Error::InvalidGltf;
 			}
 			std::uint64_t geometryNode;
-			if (error = geometryObject["node"].get_uint64().get(geometryNode); error == SUCCESS) {
-				collider.geometry.node = static_cast<std::size_t>(geometryNode);
+			if (error = geometryObject["mesh"].get_uint64().get(geometryNode); error == SUCCESS) {
+				collider.geometry.mesh = static_cast<std::size_t>(geometryNode);
```
spnda: wouldn't it be better to rename the variable `geometryMesh`?

include/fastgltf/types.hpp
```diff
@@ -2064,7 +2064,7 @@ namespace fastgltf {
         /**
          * The index of a glTF `node` which provides a mesh representation of the geometry
          */
```
spnda: The comment should also be adjusted

spnda (changes requested):

src/fastgltf.cpp
DethRaid: good call

include/fastgltf/types.hpp
DethRaid: fixed

spnda:
Sorry for responding so late.

I wanted to check the state of the extension spec and could only find this: https://github.com/KhronosGroup/glTF/pull/2424

However, in the schemas in that PR the field is still called `node` and not `mesh`. Was it said somewhere that this is guaranteed to change?

spnda:
Apparently this person was quicker and implemented the same thing, but never PR'd so: https://github.com/khiner/fastgltf/commit/95be9854030ad24e13c714226cd113f090b54645

That commit links the gltf_Physics commit that changes this, so I trust you that this is correct.

spnda (approved):

PR #153, opened 2026-09-17
Fix export directory preparation for relative paths

### LINKED ISSUE #152: Export fails when exporting to the current directory
Using `FileExporter::writeGltfJson` as an example, `"model.gltf"` results in `InvalidPath` while `./model.gltf` results in `FailedWritingFiles`.

```c++
fg::Error fg::FileExporter::writeGltfJson(const Asset& asset, const fs::path& target, const ExportOptions _options) {
	if (std::error_code ec; !fs::exists(target.parent_path(), ec) || ec) {
		fs::create_directory(target.parent_path(), ec);
		if (ec) {
			return Error::InvalidPath;
		}
	}
```

For `"model.gltf"`, `target.parent_path()` is empty. `fs::exists` returns `false` so  `fs::create_directory` is called.
For an empty path `fs::create_directory` sets an error code which causes `writeGltfJson` to return `InvalidPath`.

For `"./model.gltf"` this check succeeds because `"."` exists.
However, `writeGltfJson` later calls `writeFiles`, which in turns calls `writeFile`:

```c++
	if (!writeFiles(asset, result, target.parent_path())) {
		return Error::FailedWritingFiles;
	}
```

```c++
	bool writeFile(const DataSource& dataSource, const fs::path& baseFolder, const fs::path& filePath) {
		// Get the final normalized path. TODO: Perhaps move these filesystem checks to the parent function?
		auto finalPath = (baseFolder / filePath).lexically_normal();
		if (std::error_code ec; !fs::exists(finalPath.parent_path(), ec) || ec) {
			// If the parent folder of the destination file does not exist, we'll create it.
			fs::create_directory(finalPath.parent_path(), ec);
			if (ec) {
				return false;
			}
		}
```

`(baseFolder / filePath).lexically_normal()` normalizes `"./model.bin"` to `"model.bin"`, so `finalPath.parent_path()` becomes empty again.
This ultimately causes `writeFile`/`writeFiles` to return false then `writeGltfJson` to return `FailedWritingFiles`.

Also, `fs::create_directories` seems more appropriate than `fs::create_directory` here, since it can create missing parent directories recursively.

zhehangd:
Fixes https://github.com/spnda/fastgltf/issues/152


spnda:
I like the usage of create_directories to create them recursively. However, I'm not sure I understand why this fixes the issue? Doesn't this now assume that "." just exists and remove the sanity checks for existing directories? Isn't the actual problem with the existing code that it cannot handle paths where using parent path resolves to nothing or just `"."`? Wouldn't the cleaner fix to resolve the input path to an absolute path at the start, or am I misunderstanding something?

So make a call to `weakly_canonical` as the first line to writeGltfX() so that the code down the line can be more readable since the sanity checks for parent directories or individual files read more sensibly? Or is there anything else that that wouldn't fix?

zhehangd:
Here is a behavior cheatsheet of `exists` and `create_directories`

`exists`:
1. `dir` is non-empty and exists => returns `true`, no error
2. `dir` is non-empty and does not exist => returns `false`, no error
3. `dir` is empty => returns `false`, **no error**
4. Various filesystem errors (I'm not sure what all of them are) => returns `false` with an error

`create_directories`:
1. `dir` is non-empty and already exists => returns `false`, **no error**
2. `dir` is non-empty and does not exist => returns `true`, no error
3. `dir` is empty => returns `false` with an **error**
4. Other filesystem errors => returns `false` with an error

#### A
`create_directories` and `exists` do not behave the same way for an empty path. `exists` does not treat an empty path as an error, while `create_directories` does.

Therefore, in the existing code, an empty path — including an empty parent path provided by the user or an empty parent path produced by `lexically_normal` — will result in an error when passed to `create_directories`.

In the PR, we explicitly handle the empty-path case before calling `create_directories`.

Therefore, the PR fixes the issue.

#### B

We do not need to call `exists` before `create_directories` to check whether the directory already exists. If the directory already exists, `create_directories` does not treat that as an error. From our perspective, the outcome is effectively the same as when the directory does not exist and is created successfully: we have prepared the directory with no error.

The only difference is the return value of `create_directories`, but we do not use that return value to determine whether an error occurred.

Strictly speaking, this is not part of the “fix” itself, but the behavior is correct and it makes the code simpler.

